PR #3, opened 2016-04-28
small correction in Portuguese translation

ppaulojr:
Care Team is better translated as _"Equipe de Tratamento"_,  _"do"_ in this case would be **one** specific treatment.

Other option would be _ "Equipe de Cuidados"_


umerkhan-apple:
All localization PRs were reviewed Apple's Localization team. Any suggestions that seemed appropriate were incorporated in CK 1.1.

PR #10, opened 2016-04-29
Fix connect cell monogram truncation

stevemoser:
When an emoji is used in monogram label in the connect TVC sometimes the monogram becomes truncated. Turning on adjustsFontSizeToFitWidth is a quick fix but other fixes may be desired. I believe the phone and contacts apps on iOS display a default avatar image when an emoji is the first character in a contact's first or last name. An emoji might be a silly example but other unicode glyphs could become truncated as well. See the before and after images.

truncated-monogram

fixed-monogram

### DIFF
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -103,7 +103,12 @@ static const CGFloat ImageViewSize = 40.0;
     } else {
         _imageView.image = nil;
         _imageView.backgroundColor = OCKSystemGrayColor();
-        _monogramLabel.text = self.contact.monogram;
+        
+        NSString *monogram = self.contact.monogram;
+        NSRange stringRange = {0, MIN([monogram length], 2)};
+        stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
+        NSString *clippedMonogram = [monogram substringWithRange:stringRange];
+        _monogramLabel.text = clippedMonogram;
     }
     
     _nameLabel.text = self.contact.name;
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -85,6 +85,7 @@ static const CGFloat ImageViewSize = 40.0;
         _monogramLabel.textColor = [UIColor whiteColor];
         _monogramLabel.textAlignment = NSTextAlignmentCenter;
         _monogramLabel.font = [UIFont boldSystemFontOfSize:16.0];
+        _monogramLabel.adjustsFontSizeToFitWidth = YES;
         [self addSubview:_monogramLabel];
     }
     
--- a/CareKit/Connect/OCKContact.h
+++ b/CareKit/Connect/OCKContact.h
@@ -137,8 +137,7 @@ OCK_CLASS_AVAILABLE
 /**
  A string indicating the monogram for a contact.
  
- If a monogram is not provided, the image will be used for the contact.
- If a monogram is available, it will be clipped to two glyphs.
+ The monogram will be clipped to two glyphs.
  */
 @property (nonatomic, readonly) NSString *monogram;
 
--- a/CareKit/Connect/OCKContact.h
+++ b/CareKit/Connect/OCKContact.h
@@ -138,6 +138,7 @@ OCK_CLASS_AVAILABLE
  A string indicating the monogram for a contact.
  
  If a monogram is not provided, the image will be used for the contact.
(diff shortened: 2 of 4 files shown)


umerkhan-apple:
@stevemoser Thanks for this PR! 

I think we could combine `adjustsFontSizeToFitWidth` along with clipping the string length to be 2 characters. This is to prevent the monogram from becoming a really small font size if a long string is provided. What do you think?

CareKit/Connect/OCKConnectTableViewCell.m
```diff
@@ -102,7 +103,12 @@ - (void)updateView {
     } else {
         _imageView.image = nil;
         _imageView.backgroundColor = OCKSystemGrayColor();
-        _monogramLabel.text = self.contact.monogram;
+        
+        NSString *monogram = self.contact.monogram;
+        NSRange stringRange = {0, MIN([monogram length], 2)};
+        stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
+        NSString *clippedMonogram = [monogram substringWithRange:stringRange];
+        _monogramLabel.text = clippedMonogram;
```
stevemoser: @umerkhan-apple Sounds good. Were you thinking of clipping it like this?

umerkhan-apple: @stevemoser Same idea, but with a slightly different approach.
```
NSString *monogram = self.contact.monogram;
NSString *clippedMonogram = [monogram substringToIndex:MIN(monogram.length, 2)];
```

We should also update this in `OCKConnectTableViewHeader.m`

stevemoser: @umerkhan-apple Will do though I'll be gone for a few days so I will get to it when I get back. However while subStringToIndex is simpler it breaks surrogate pairs. Is the simplify worth it in this case? When I use subStringToIndex with my example above the emoji becomes a half 'A' in a box unicode character (sorry I don't know the exact name).

matteocrippa:
Isn't better that monogram is auto-generated according the name provided by the user?
I added an on boarding page to provide physician details in my app and monogram is used to be generated with this string extension:

```
extension String {
    func getMonogram() -> String {
        var tmp = self.componentsSeparatedByString(" ")
        var name = ""
        var surname = ""
        name = String(tmp[0].characters.first!)
        if tmp.count > 1 {
            surname = String(tmp[1].characters.first!)
        }
        return name+surname
    }
}
```

umerkhan-apple:
@matteocrippa I really like that idea! 
@stevemoser I agree that we don't have half of the unicode character in the monogram. How do you feel about auto-generating the monogram?

oliverschaefer:
@matteocrippa @umerkhan-apple I like the idea of auto-generated monograms. When implementing, please keep in mind that the physician's name might contain (academic) titles. Some regions (e.g. Europe) place them in front of the given name. E.g. John Appleseed, M.D. (USA) = Dr. John Appleseed (Germany).

matteocrippa:
@oliverschaefer that part it would be tricky, we probably have two way to approach that: blacklisting or check for dots inside the word. Let's talk about that in #21

umerkhan-apple:
@stevemoser Let's proceed with this PR without adding the auto-generated monograms, as that will be added in #21.

umerkhan-apple:
@stevemoser Could you resolve merge conflicts? PR looks good to be merged otherwise. 👍

stevemoser:
@umerkhan-apple 
- [x] Updated OCKConnectTableViewHeader.m to adjust font size to fit
- [x] Move clipping logic to OCKContact
- [x] Updated header for OCKContact

Anything else?

CareKit/Connect/OCKContact.h
```diff
@@ -138,6 +138,7 @@ OCK_CLASS_AVAILABLE
  A string indicating the monogram for a contact.
  
  If a monogram is not provided, the image will be used for the contact.
```
umerkhan-apple: Remove this line.

CareKit/Connect/OCKContact.h
```diff
@@ -138,6 +138,7 @@ OCK_CLASS_AVAILABLE
...
+ If a monogram is available, it will be clipped to two glyphs.
  */
```
umerkhan-apple: I think it should just say "The monogram will be clipped to two glyphs."

umerkhan-apple:
@stevemoser Minor nits. Apart from that, it is good to go! 👍

stevemoser:
@umerkhan-apple And done. 👍

umerkhan-apple:
@stevemoser Looks great! Merging.

PR #17, opened 2016-04-30
Added colon

akkatracker:
Added a colon where it was necessary to match the styling of the following paragraph


umerkhan-apple:
r=m3 👍

PR #18, opened 2016-04-30
Tint issues

matteocrippa:
A couple of fixes about tint connected issues discovered in OCKConnectVC and OCKSymptomVC

### DIFF
--- a/CareKit/Connect/OCKConnectViewController.m
+++ b/CareKit/Connect/OCKConnectViewController.m
@@ -219,9 +219,11 @@
     OCKContact *contact = _sectionedContacts[indexPath.section][indexPath.row];
     
     OCKConnectDetailViewController *detailViewController = [[OCKConnectDetailViewController alloc] initWithContact:contact];
+    detailViewController.view.tintColor = self.view.tintColor;
     detailViewController.delegate = self.delegate;
     detailViewController.masterViewController = self;
     detailViewController.showEdgeIndicator = _showEdgeIndicators;
+    self.navigationController.navigationBar.tintColor = self.view.tintColor;
     [self.navigationController pushViewController:detailViewController animated:YES];
     
     [tableView deselectRowAtIndexPath:indexPath animated:YES];
--- a/CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
+++ b/CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
@@ -106,6 +106,7 @@
 
     NSAssert(self.navigationController, @"OCKSymptomTrackerViewController must be embedded in a navigation controller.");
     _weekViewController.symptomTrackerWeekView.delegate = self;
+    [_tableView reloadData];
 }
 
 - (void)showToday:(id)sender {
--- a/CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
+++ b/CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
@@ -84,8 +84,7 @@
                                                                               style:UIBarButtonItemStylePlain
                                                                              target:self
                                                                              action:@selector(showToday:)];
-    self.navigationItem.rightBarButtonItem.tintColor = self.progressRingTintColor;
-    
+
     _tableView = [[UITableView alloc] initWithFrame:CGRectZero style:UITableViewStylePlain];
     _tableView.dataSource = self;
     _tableView.delegate = self;
@@ -103,7 +102,8 @@
 
 - (void)viewDidAppear:(BOOL)animated {
     [super viewDidAppear:animated];
-    
+    self.navigationItem.rightBarButtonItem.tintColor = self.view.tintColor;
+
     NSAssert(self.navigationController, @"OCKSymptomTrackerViewController must be embedded in a navigation controller.");
     _weekViewController.symptomTrackerWeekView.delegate = self;
 }


CareKit/Connect/OCKConnectViewController.m
```diff
@@ -219,9 +219,11 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
     OCKContact *contact = _sectionedContacts[indexPath.section][indexPath.row];
     
     OCKConnectDetailViewController *detailViewController = [[OCKConnectDetailViewController alloc] initWithContact:contact];
+    detailViewController.view.tintColor = self.view.tintColor;
     detailViewController.delegate = self.delegate;
     detailViewController.masterViewController = self;
     detailViewController.showEdgeIndicator = _showEdgeIndicators;
+    self.navigationController.navigationBar.tintColor = self.view.tintColor;
```
umerkhan-apple: Would we want to apply this to other view controllers (such as CareCard, Symptom Tracker, Insights)?

CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
```diff
@@ -84,8 +84,7 @@ - (void)viewDidLoad {
                                                                               style:UIBarButtonItemStylePlain
                                                                              target:self
                                                                              action:@selector(showToday:)];
-    self.navigationItem.rightBarButtonItem.tintColor = self.progressRingTintColor;
-
```
umerkhan-apple: The developer can provide a progressRingTintColor that is then applied to the ring and the `rightBarButtonItem`. We want this for consistency in the color scheme. If the developer chooses to use a different color than the progress ring for the `rightBarButtonItem`, then they can just specify it on the navigation controller.

We probably don't want to remove this functionally.

This behavior is also mimicked in `OCKCareCardViewController`.

umerkhan-apple:
@matteocrippa Minor feedback. Let me know how you think we should proceed!

CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
matteocrippa: Idea, isn't better that tintColor first override the overall colorscheme and the, only if applied, progress override the right button?

CareKit/Connect/OCKConnectViewController.m
matteocrippa: Yep, I think that we should do that. If I remember well they do the same in RK.
Let's figure out how to approach the progress and the 'll provide a new PR

CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
umerkhan-apple: If I understand your suggestion correctly, I believe that is how the current implementation works.
The `rightBarButtonItem` adopts the tintColor for the overall color scheme unless a `progressRingTintColor` is applied.

matteocrippa: Yes, I confirm you have understood in the right manner, but at this very moment it seems that rightBarButtonItem stays always with the default tintColor (blue) with the current implementation in master repository, I was forced to tweak the code like PR above to have it working.

umerkhan-apple: I am unable to reproduce this issue.
In `OCKSample`, a red tint color is applied to the entire window.
This custom tint color can be seen in the Symptom Tracker and the Connect Detail View Controller.

matteocrippa: @umerkhan-apple you can reproduce using this [branch](https://github.com/matteocrippa/CareKit/tree/tintColorIssue) , as you can see the only component that goes purple is the name of the day, nor Today button or other tint Object are purple as expected.

simulator screen shot 4 may 2016 20 12 55

umerkhan-apple:
@matteocrippa When I run the sample app using your branch, I get the following:

simulator screen shot may 4 2016 12 21 36 pm

Then I change the `progressRingTintColor` to purple, and I get the following:

simulator screen shot may 4 2016 12 23 45 pm

Are you using the latest master? If so, could provide exact steps to reproduce the condition shown in your screenshot. Thanks! :)

CareKit/SymptomTracker/OCKSymptomTrackerViewController.m
umerkhan-apple: @matteocrippa When I run the sample app using your branch, I get the following:

simulator screen shot may 4 2016 12 21 36 pm

Then I change the `progressRingTintColor` to purple, and I get the following:

simulator screen shot may 4 2016 12 23 45 pm

Are you using the latest master? If so, could provide exact steps to reproduce the condition shown in your screenshot. Thanks! :)

umerkhan-apple: One thing I noticed is that you are changing the view's tint color, not the progressRingTintColor. 

The progressRingTintColor sets the tintColor for the Ring, the background of the Day Label, and the Today button.

After setting that, if the dev wishes to change the Today's tintColor, they can do so by setting the tintColor on the navigationItem. Although, the tintColor for the progressRing and the dayLabel will always be the same.

matteocrippa: Yes, I'm using the latest master, and I confirm you that setting ```progressRingTintColor``` it works as shown in your screenshot.

But if you instead set only ```tintColor``` you will notice that the Today button will stay red instead of purple, and only the selected day background goes purple.

matteocrippa: Yes, I confirm you I'm using the tintColor instead, I managed to do so because in RK I was using that var to get rid of the default color of some elements.

Probably instead using my PR, we can improve the [docs](http://carekit.org/docs/Classes/OCKSymptomTrackerViewController.html#//api/name/progressRingTintColor) adding this note about ```progressRingTintColor```, what do you think?

umerkhan-apple: That's a good idea! Let's proceed with that.

PR #14, opened 2016-04-29
Unify code blocks.

cyhsutw:
Unify code blocks of documents.

Summary:
1. replace tab-based code blocks with backtick ones (fenced code blocks)
2. add syntax highlighting for Swift and Objective-C
3. unify whitespace and indentions in code blocks
4. remove trailing whitespace

PR #20, opened 2016-05-02
[Dcumentation] Fix misplaced unicode chars "\u{0B0C}" and "\u{0B0F}"

cyhsutw:
Hi all,

I converted some example code to Swift in #14.
However, some characters is not displaying correctly due to a bug made by me 😭 

We'd like to display `°F` and `°C` on the screen, but I put the `C` and `F` into the hexadecimal block the the text was rendered as  `ଏ` (0B0F) and `ଌ` (0B0C).

I've fixed this. Really sorry about the bug :bow:
 
It would be appreciated if you could take some time reviewing the changes.

### DIFF
--- a/docs/AccessingCarePlanData/AccessingCarePlanData-template.md
+++ b/docs/AccessingCarePlanData/AccessingCarePlanData-template.md
@@ -250,8 +250,8 @@ func symptomTrackerViewController(viewController: OCKSymptomTrackerViewControlle
                             quantitySample: sample,
                             quantityStringFormatter: nil,
                             unitStringKeys: [
-                                HKUnit.degreeFahrenheitUnit() : "\u{00B0F}",
-                                HKUnit.degreeCelsiusUnit()    : "\u{00B0C}"
+                                HKUnit.degreeFahrenheitUnit() : "\u{00B0}F", // °F
+                                HKUnit.degreeCelsiusUnit()    : "\u{00B0}C"  // °C
                             ],
                             userInfo: nil
                         )


umerkhan-apple:
@cyhsutw No worries. Thanks for the quick update! r=m3, 👍

cyhsutw:
@umerkhan-apple Thanks for the rapid response!

docs/AccessingCarePlanData/AccessingCarePlanData-template.md
```diff
@@ -250,8 +250,8 @@ func symptomTrackerViewController(viewController: OCKSymptomTrackerViewControlle
                             quantitySample: sample,
                             quantityStringFormatter: nil,
                             unitStringKeys: [
-                                HKUnit.degreeFahrenheitUnit() : "\u{00B0F}",
-                                HKUnit.degreeCelsiusUnit()    : "\u{00B0C}"
+                                HKUnit.degreeFahrenheitUnit() : "\u{00B0}F", // °F
```
yas375: How about actually using `°` instead of `\u{00B0}`? Any reason not to?

cyhsutw: Hi @yas375 Thanks for feedback!
I was thinking putting the `°` directly too. 

I think the reason for using escaped hexadecimal is that it's hard to type or copy-paste unicode chars in some environments. 

However, since it's an iOS app (and will be built using Xcode on OS X), everyone should be able to type/display chars.

I'm +1 with using `°` directly.

@umerkhan-apple What do you think?

PR #21, opened 2016-05-03
Monogram

matteocrippa:
Added support for auto monogram generation according #10 
I ported my func in objc and it picks the first and last word inside the name, so I removed also the assert for monogram and image not to be nil.

The only issue I figured out is that if you put a title in front of the name it will pick that one.

Eg.

Dr. Maria Ruiz will be DR and not MR anymore.

By the way the func check for monogram and activate the generator only it is nill.

Dunno if is better to put a check for name to be forced now not to be nil.

### DIFF
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -103,7 +103,12 @@ static const CGFloat ImageViewSize = 40.0;
     } else {
         _imageView.image = nil;
         _imageView.backgroundColor = OCKSystemGrayColor();
-        _monogramLabel.text = self.contact.monogram;
+        
+        NSString *monogram = self.contact.monogram;
+        NSRange stringRange = {0, MIN([monogram length], 2)};
+        stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
+        NSString *clippedMonogram = [monogram substringWithRange:stringRange];
+        _monogramLabel.text = clippedMonogram;
     }
     
     _nameLabel.text = self.contact.name;
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -105,7 +105,12 @@ static const CGFloat ImageViewSize = 40.0;
         _monogramLabel.text = self.contact.monogram;
     }
     
-    _nameLabel.text = self.contact.name;
+    if (self.contact.title) {
+        _nameLabel.text = [NSString stringWithFormat:@"%@ %@",self.contact.title, self.contact.name];
+    } else {
+        _nameLabel.text = self.contact.name;
+    }
+    
     _relationLabel.text = self.contact.relation;
 }
 
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -101,7 +101,6 @@ static const CGFloat ImageViewSize = 40.0;
         _monogramLabel.text = nil;
     } else {
         _imageView.image = nil;
-<<<<<<< cbbd275cb2b82d1333c5da03a3cd207a09442c16
         _imageView.backgroundColor = OCKSystemGrayColor();
         
         NSString *monogram = self.contact.monogram;
@@ -109,9 +108,6 @@ static const CGFloat ImageViewSize = 40.0;
         stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
         NSString *clippedMonogram = [monogram substringWithRange:stringRange];
         _monogramLabel.text = clippedMonogram;
-=======
-        _monogramLabel.text = self.contact.monogram;
->>>>>>> Better highlight state for connect cell monogram.
     }
     
     [self updateImageViewBackgroundColor];
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -107,7 +107,6 @@ static const CGFloat ImageViewSize = 40.0;
     
     _nameLabel.text = self.contact.name;
     
-    
     _relationLabel.text = self.contact.relation;
 }
 
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -113,7 +113,6 @@ static const CGFloat ImageViewSize = 40.0;
     [self updateImageViewBackgroundColor];
     
     _nameLabel.text = self.contact.name;
-    
     _relationLabel.text = self.contact.relation;
 }
 
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -101,13 +101,7 @@ static const CGFloat ImageViewSize = 40.0;
         _monogramLabel.text = nil;
     } else {
         _imageView.image = nil;
-        _imageView.backgroundColor = OCKSystemGrayColor();
-        
-        NSString *monogram = self.contact.monogram;
-        NSRange stringRange = {0, MIN([monogram length], 2)};
-        stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
-        NSString *clippedMonogram = [monogram substringWithRange:stringRange];
-        _monogramLabel.text = clippedMonogram;
+        _monogramLabel.text = self.contact.monogram;
     }
     
     [self updateImageViewBackgroundColor];
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -85,6 +85,7 @@ static const CGFloat ImageViewSize = 40.0;
         _monogramLabel.textColor = [UIColor whiteColor];
         _monogramLabel.textAlignment = NSTextAlignmentCenter;
         _monogramLabel.font = [UIFont boldSystemFontOfSize:16.0];
+        _monogramLabel.adjustsFontSizeToFitWidth = YES;
         [self addSubview:_monogramLabel];
     }
     
--- a/CareKit/Connect/OCKConnectTableViewCell.m
+++ b/CareKit/Connect/OCKConnectTableViewCell.m
@@ -98,10 +98,10 @@ static const CGFloat ImageViewSize = 40.0;
     
     if (self.contact.image) {
         _imageView.image = self.contact.image;
-        _imageView.backgroundColor = [UIColor clearColor];
         _monogramLabel.text = nil;
     } else {
         _imageView.image = nil;
+<<<<<<< cbbd275cb2b82d1333c5da03a3cd207a09442c16
         _imageView.backgroundColor = OCKSystemGrayColor();
         
         NSString *monogram = self.contact.monogram;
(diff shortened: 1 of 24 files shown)


matteocrippa:
Improved the way monogram is generated, now it makes use of a blacklist array where all words with at least a dot inside are written, this avoid the issue pointed out by @oliverschaefer in #10

oliverschaefer:
@matteocrippa Maybe it will be clearer and easier if we add a title property to `OCKContact` that can hold the academic title (if any). Let's see what @umerkhan-apple says.

p2:
For the blacklist, it's noteworthy that the dot-approach will also skip truncated middle names, e.g. the "A." in "James A. Corey". I guess it's not really a problem that this will become "JC" instead of "JAC".

But note that “MD” (or "PhD") is often spelled without the dots, so this approach will not skip those titles in this case.

matteocrippa:
we can try upgrading with an extra list of blacklisted words like MD or PhD, but we have also to localise it in each country

p2:
It could also be "MD-PhD" or "MD, PhD" or even crazier stuff like "MD, MBA, FACMI". But usually, these titles appear after a comma, so maybe just ignore everything after a comma if more than one name/word was present before the comma?

oliverschaefer:
Wouldn't a dedicated title property be easier especially in terms of localization?

Von meinem iPhone gesendet

matteocrippa:
Prob the best way is to split title in a new field, and have auto-monogram to generate it using only name, surname.

This will avoid each possible issue about blacklisting etc.

matteocrippa:
@umerkhan-apple just improved the overall structure, as suggested by @oliverschaefer I pushed a new PR with a new var that will store (if exists) the academic title of the contact.
The monogram generator is still in place also.

p2:
I think this is being over-engineered. There's no need for a title property, all you need to know is who the person on the care team is, you likely already know the person and the title can be pre/appended to the name or be omitted entirely.

As for the monogram, you most likely populate `OCKContact` from another contact element, which probably has separate fields for prefixes, suffixes, family- and given names. It's easier to generate a correct monogram from your original source than to do it inside CareKit from the full name string, so I'm not 100% convinced of the use case here either.

oliverschaefer:
@p2 Thanks for your input. I don't think that an optional title field would instantly be an over-engineering. But as usual I'm open for discussion. What do you think @umerkhan-apple ?

p2:
Thing is it won't work they way it's constructed right now, it assumes in two places to display `"{title} {name}"`, which is wrong for all of North America. In the Americas and Europe you can have titles before and after the name. In Asian countries it usually follows the name but can even follow the given name.

So you must at least store prefixes and suffixes in addition to a full name string, or you just keep the one “name” string. I think the latter is preferable, just assign one string like it should show up. Hell this could even be _“Dr. Bob”_ if it's your PCP.

matteocrippa:
It's easier to set an ```enum``` like:

```
enum TitlePosition {
case .None,
case .Prefix,
case .Suffix
}
```

 this will allow developer to manage this in the better way according localisation

oliverschaefer:
@p2 You are right, we cannot append the title to the name (or the other way around). The easiest thing would be to display the title separately (e.g. as a detailTextLabel in a UITableViewCell). That would fulfill the demand of entering a title (which some of my physician-colleagues are quite strict about) and also keep things easy in terms of localization and monogram generation (which this topic originally was all about ;-) )

p2:
I still maintain having only a `name` property that includes all titles is by far the best solution for CareKit.

oliverschaefer:
@p2 Probably yes, but killing the monogram auto-generation feature.

matteocrippa:
@p2 @oliverschaefer at this very moment it acts like iOS ABAddressBook:
- you can set a Name and Surname
- you can set a Title

But it generates the monogram without the title and append it before the name

img_1858

img_1857

p2:
@oliverschaefer Ignoring everything ending with a period and everything after the first comma – probably only if there are at least two words before the comma – could still result in good monogram auto-creation. For edge cases the app developer can just create the monogram herself.

umerkhan-apple:
@matteocrippa If the auto-generation will have edge cases that cannot be handled by the implementation, then it might be best to just ask the developer for the monogram. 

Currently, the monogram is non-null, we could potentially make it null_resettable. If the developer does not provide one, we can auto-generate one for them.

@p2 @oliverschaefer Let me know how you guys feel about this approach.

p2:
I feel making it `null_resettable` and auto-generating one if it's null with @matteocrippa 's approach would be reasonable. There will be edge cases, then it's up to the developer to work around those to improve her/his app.

oliverschaefer:
Agree

Von meinem iPhone gesendet

matteocrippa:
I was looking at [CNContact](https://developer.apple.com/library/ios/documentation/Contacts/Reference/CNContact_Class/index.html#//apple_ref/occ/instp/CNContact/namePrefix) that will be for sure in some cases the real source of the data the user is going to provide.

You can see there that there are two vars, one for **nameSuffix** and one for **namePrefix** why not approach it in this way or (probably) even better use ```CNContact``` instance or reference and add only the missing fields to the data model (e.g. relation, tintColor)?

In this way monogram will be generated easily looking at **.givenName** and **.familyName** and the user will also be able to easily store the information inside his/her local address book too.

matteocrippa:
b8b8b0a provides monogram as ```null_resettable```

p2:
@matteocrippa That is what I meant, structured names usually have prefixes and suffixes. I don't think it makes sense to have yet another structured way of maintaining contact names in CareKit. But, often, the source of the name is structured, either `CNContact` or some API return. From there it's easy to create a monogram because you only look at given, middle and family names. And if you don't create a monogram, the suggestions posited above are good enough IMHO.

umerkhan-apple:
@matteocrippa I think we should take out the `title` property from `OCKContact`. A developer can use a formatter to create the full name string and provide it to the `OCKContact` object. Though we should keep the `null_resettable` on the monogram. We also won't need a blacklist, because if the developer would like a monogram other than the auto-generated one, they can specify one.

matteocrippa:
@umerkhan-apple pushed new commits, now it should work as you suggested

CareKit/Connect/OCKContact.m
```diff
@@ -134,4 +133,37 @@ - (instancetype)copyWithZone:(NSZone *)zone {
     return contact;
 }
 
+#pragma mark - Monogram
```
umerkhan-apple: add new line

Sample/OCKSample/SampleData.swift
```diff
@@ -57,67 +57,67 @@ class SampleData: NSObject {
             phoneNumber: CNPhoneNumber(stringValue: "888-555-5512"),
             messageNumber: CNPhoneNumber(stringValue: "888-555-5512"),
             emailAddress: "mruiz2@mac.com",
-            monogram: "MR",
```
umerkhan-apple: Lets keep this one to test both use-cases.

CareKit/Connect/OCKConnectTableViewCell.m
```diff
@@ -106,6 +106,8 @@ - (void)updateView {
     }
     
     _nameLabel.text = self.contact.name;
+    
+
```
umerkhan-apple: Remove extra new lines.

umerkhan-apple:
@matteocrippa Minor nits and merge conflicts. Apart from that, looks good to go! 👍

CareKit/Connect/OCKContact.m
```diff
@@ -134,4 +133,37 @@ - (instancetype)copyWithZone:(NSZone *)zone {
...
+- (void)setMonogram:(NSString *)monogram {
+    if (!monogram) {
+        monogram = [self generateMonogram:_name];
```
p2: Could this be implemented on the getter, not the setter? Then the `self.monogram` "hack" in the initializer is not needed.

matteocrippa:
updated my fork with master and merged in my branch

CareKit/Connect/OCKConnectTableViewCell.m
```diff
@@ -101,12 +101,19 @@ - (void)updateView {
...
     } else {
         _imageView.image = nil;
-        _monogramLabel.text = self.contact.monogram;
+        _imageView.backgroundColor = OCKSystemGrayColor();
+        
+        NSString *monogram = self.contact.monogram;
+        NSRange stringRange = {0, MIN([monogram length], 2)};
+        stringRange = [monogram rangeOfComposedCharacterSequencesForRange:stringRange];
+        NSString *clippedMonogram = [monogram substringWithRange:stringRange];
+        _monogramLabel.text = clippedMonogram;
```
umerkhan-apple: PR https://github.com/carekit-apple/CareKit/pull/10 takes care of the clipping as part of `OCKContact`.

CareKit/Connect/OCKConnectTableViewCell.m
```diff
@@ -101,12 +101,19 @@ - (void)updateView {
...
     }
     
     [self updateImageViewBackgroundColor];
     
     _nameLabel.text = self.contact.name;
+
```
umerkhan-apple: Remove extra line.

CareKit/Connect/OCKConnectTableViewCell.m
umerkhan-apple: PR https://github.com/carekit-apple/CareKit/pull/10 takes care of the monogram clipping as part of `OCKContact`.

CareKit/Connect/OCKContact.m
```diff
@@ -49,8 +49,7 @@ - (instancetype)initWithContactType:(OCKContactType)type
                        emailAddress:(NSString *)emailAddress
                            monogram:(NSString *)monogram
                               image:(UIImage *)image {
-    NSAssert((monogram || image), @"An OCKContact must have either a monogram or an image.");
-    
+
```
umerkhan-apple: Remove extra line.

CareKit/Connect/OCKContact.m
```diff
@@ -143,4 +133,44 @@ - (instancetype)copyWithZone:(NSZone *)zone {
...
+- (void)setMonogram:(NSString *)monogram {
+    if (!monogram) {
+        monogram = [self generateMonogram:_name];
+    }
+    _monogram = [monogram copy];
+}
+
+- (NSString *)generateMonogram:(NSString *)name {
+
```
umerkhan-apple: Remove extra line.

matteocrippa:
sorry, now it should be fixed

matteocrippa:
hey @umerkhan-apple this PR is still pending, are you considering to merge or do you prefer to change the approach for monogram?

Sample/OCKSample/SampleData.swift
```diff
@@ -59,65 +59,65 @@ class SampleData: NSObject {
...
             messageNumber: CNPhoneNumber(stringValue: "888-555-5512"),
             emailAddress: "billjames2@mac.com",
-            monogram: "BJ",
+            monogram: nil,
             image: nil),
-        
+
         OCKContact(contactType: .Personal,
-            name: "Tom Clark",
+            name: "Tom Van de Clark",
```
umerkhan-apple: Can we remove Van de?

umerkhan-apple:
@matteocrippa Apart from one minor nit. This looks good!
@YuanZhu-apple Can you take a look as well?

Sample/OCKSample/SampleData.swift
matteocrippa: I put that only to check the issue w/ multiple words inside an user name, but we can for sure remove it.

umerkhan-apple: That makes sense. Let's go ahead and put move the check into the test app/

umerkhan-apple:
@YuanZhu-apple Friendly ping! :)

PR #23, opened 2016-05-03
Small UI Improvements

brynbodayle:
Some small improvements to usability and visuals. See commit comments for further detail.

Since these were rather small I didn't feel they merited an issue, but happy to open one if that's preferred.


umerkhan-apple:
Looks great! r=m3!

PR #26, opened 2016-05-04
Nepali localization

bishalg:
Nepali localization string added


yas375:
I would suggest to squash all commits into one and to remove changes from `Sample/OCKSample.xcodeproj/project.pbxproj` and to `Sample/OCKSample/RootViewController.swift` as they don't seem to be related to the pull request's subject.

bishalg:
@yas375 : I have send another pull request with only Localization change 
https://github.com/carekit-apple/CareKit/pull/49
@umerkhan-apple : check #49

PR #28, opened 2016-05-04
Increased tap target size in week view control.

brynbodayle:
I increased the size of the buttons for the week view in the Care Card and Symptom Tracker so that you can tap on the day of the week label to change the day as well as by tapping the circle or heart underneath. This doesn't change anything visually.


umerkhan-apple:
Awesome! Thanks for this enhancement! 👍

PR #31, opened 2016-05-06
Add a (read only) property encapsulating whetner an `OCKCarePlanEvent` is available

apbendi:
This adds a read only property that uses the `OCKCarePlanEvent`'s state, and the `resultResettable` property of its activity to expose whether the event is available to the user. It seems useful to provide this definition internally; I'd imagine most every CareKit app would have to perform this check.

### DIFF
--- a/CareKit/CarePlan/OCKCarePlanEvent.h
+++ b/CareKit/CarePlan/OCKCarePlanEvent.h
@@ -96,6 +96,12 @@ OCK_CLASS_AVAILABLE
 @property (nonatomic, readonly) OCKCarePlanEventState state;
 
 /**
+ The availability of this event, based on its `state`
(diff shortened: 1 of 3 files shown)


umerkhan-apple:
Looks good to me! @YuanZhu-apple could you please review?

umerkhan-apple:
@YuanZhu-apple Friendly ping.

YuanZhu-apple:
@apbendi This framework is designed very open, developer has the freedom to decide the behavior when the state is `StateNotCompleted`. I think it is make sense for you implement this as a convenient function in your app, but not quite necessary for this framework.

apbendi:
Fair enough, @YuanZhu-apple. Thanks for reviewing!

PR #35, opened 2016-05-10
Remove superfluous characters

vtourraine:
Remove trailing `.` characters in image descriptions, causing new lines and isolated characters when rendered in HTML.

screen shot 2016-05-10 at 5 42 36 pm


umerkhan-apple:
Thanks! 👍

PR #38, opened 2016-05-12
OCKSymptomTrackerViewController doesn't update contents when activity list is changed

weijentu:
Hi there,
I noticed that ``OCKSymptomTrackerViewController`` doesn't update its contents when activity list is changed. According to ``handleActivityListChange`` in ``OCKCarePlanStore``,
I guess ``carePlanStoreActivityListDidChange`` should be correct method name
by original intention.

Cheers,


YuanZhu-apple:
@weijentu Thanks for catching this! 
R=me.

PR #43, opened 2016-05-12
Various UI Performance Enhancements

brynbodayle:
Made some improvements to the rendering speed of the insight view cells by reducing the amount of blended layers in the cells.

See before and after. These screenshots are with the "Color Blended Layers" setting enabled in the iOS Simulator.

screen shot 2016-05-11 at 6 15 59 pm
screen shot 2016-05-12 at 9 34 29 am

### DIFF
--- a/CareKit/PDF/OCKDocument.m
+++ b/CareKit/PDF/OCKDocument.m
@@ -174,7 +174,7 @@ static NSString *imageTagFromImage (UIImage *image) {
 }
 
 static NSString *imageTagFromView (UIView *view) {
-    UIGraphicsBeginImageContextWithOptions(view.frame.size, YES, 2.0);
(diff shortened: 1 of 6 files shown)


CareKit/PDF/OCKDocument.m
```diff
@@ -174,7 +174,7 @@ - (instancetype)copyWithZone:(NSZone *)zone {
 }
 
 static NSString *imageTagFromView (UIView *view) {
-    UIGraphicsBeginImageContextWithOptions(view.frame.size, YES, 2.0);
+    UIGraphicsBeginImageContextWithOptions(view.frame.size, YES, 0.0);
```
YuanZhu-apple: This means to output on a PDF file, should it be depends on the device's screen resolution?

umerkhan-apple:
Apart from @YuanZhu-apple's comment, looks good! r=me.

Issue #37, opened 2016-05-11
Adding activity with identical identifier results in odd callback parameters


apbendi:
The documentation states that it is not possible to add an activity with an identical identifier to one already stored (https://github.com/carekit-apple/CareKit/blob/master/CareKit/CarePlan/OCKCarePlanStore.h#L108).

So far, in a number of tests, it seems to me that adding an activity with an existing identifier results in a callback with `YES` for the `success` parameter **and** with an `error` populated stating an activity with the identifier already exists. No activity is actually added. This seems counter intuitive, but it's not clear to me what the expected behavior should be.

Note that any app that attempts to add an activity after the first start of the app will encounter this behavior. The sample app also depends on it. (https://github.com/carekit-apple/CareKit/blob/master/Sample/OCKSample/SampleData.swift#L94)

Diving into the implementation, unless I'm totally missing something (always possible!), the behavior does not seem to be intended. For one, the app nil checks the "inout" error parameter that was passed into the method [two](https://github.com/carekit-apple/CareKit/blob/master/CareKit/CarePlan/OCKCarePlanStore.m#L139) [times](https://github.com/carekit-apple/CareKit/blob/master/CareKit/CarePlan/OCKCarePlanStore.m#L152), when it seems the intention was to nil check the local `errorOut`.

Additionally, even if the [first check](https://github.com/carekit-apple/CareKit/blob/master/CareKit/CarePlan/OCKCarePlanStore.m#L158) was corrected, I'm not sure I understand it's purpose. It seems to me you'd want to return the [subsequent error](https://github.com/carekit-apple/CareKit/blob/master/CareKit/CarePlan/OCKCarePlanStore.m#L141) if an `item` were found, or return the value of the `errorOut` variable if something else went wrong.

I'm happy to take a stab at solving this, but want to clear up what the actual intended behavior is. Should the store return an error anytime an activity with the same identifier is added? (As far as I can tell, this would mean apps would have to take care to only add activities at first start, or when first added. This may in fact be reasonable, but I wanted to confirm given the behavior of the sample app).

YuanZhu-apple:
Should be `No` for `success`, can you make pull request? Thanks!

apbendi:
Just opened #48. I realized even I was slightly confused about **why** this was wrong when I reported this issue. **What** was wrong was still valid, though, and my PR hopefully addresses it properly. Thanks! :)

PR #48, opened 2016-05-18
Fix store callback when an activity with an existing identifier is added

apbendi:
Previously, the callback would contain `YES` and an error. Now, it properly returns `NO` and an error. I also tried to clarify the semantics to avoid future confusion, at the expense of a bit of verbosity.

As a side note, the naming here makes it very easy to get confused, which it seems is what happened to both the original implementer and to me when I reported #37. I might prefer calling the `error` parameter `errorPtr` or similar, but didn't make that change here as the same convention is used throughout the class.

### DIFF
--- a/CareKit/CarePlan/OCKCarePlanStore.m
+++ b/CareKit/CarePlan/OCKCarePlanStore.m
@@ -124,7 +124,7 @@ static NSString * const OCKAttributeNameDayIndex = @"numberOfDaysSinceStart";
     NSParameterAssert(sourceItem);
 
     NSManagedObjectContext *context = _managedObjectContext;
-    if (context == nil) {
+    if (nil == context) {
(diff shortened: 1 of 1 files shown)


apbendi:
Friendly Bump! This seems like an important bug to fix; happy to take feedback :)

umerkhan-apple:
Thanks for this PR and fix!

We like to avoid multiple returns inside a function, but I see how in this use-case the function is easier to understand. @YuanZhu-apple What do you think?

YuanZhu-apple:
r=me.

ninoguba:
Regardless of success or failure to add the activity to the store, this change always spit out the following message:

> Optional("The operation couldn’t be completed. (OCKErrorDomain error 2.)")

PR #50, opened 2016-05-19
IB_DESIGNABLE for OCKCareCardButton

bishalg:
Made OCKCareCardButton IB_DESIGNABLE 
Added IBInspectable for lineWidth and buttonSize


umerkhan-apple:
@bishalg Thanks for the pull request. Is there any reason for making this particular class public and IB_DESIGNABLE?

We would probably want to make Care Card, Symptom Tracker, Insights Dashboard, Bar Chart and Connect IB_DESIGNABLE, since those are public classes.

bishalg:
I am wondering why are so many class Private ?
By saying - "We would probably want to make" you mean soon CareKit will make them Public ? Or do you want me to make then Public ?

In my particular app I would never need OCKCareCardTableViewCell as my app has completely different UI then CareKit, I would never use ViewController provided by CareKit as its too much bind to some particular view. But for my case I would need to use lot of methods and helper functions to get my desired Insights or Report generation. 

I would need to use _NSDateComponents+CarePlan_ for my custom view controller also some common DataComponent like _ock_componentsWithDate_ which are all private and I have to manually convert them to my custom class and add and use them.

Is this a design decision of framework or am i getting it completely wrong. I feel Framework should expose those helper methods so that end user can easily use the framework for each particular case.

Thanks for your help.

bishalg:
My use case is as follows - 

screen shot 2016-05-30 at 2 51 52 pm

umerkhan-apple:
@bishalg Is this report being generated to a PDF or is it a view controller inside the app?

bishalg:
@umerkhan-apple : 

1.  Please can you explain certain things being private in CareKit. Like I said before - _NSDateComponents+CarePlan_ and _ock_componentsWithDate_ were among the few I need in my project.

2.  In this particular case its a Controller which is then used to generate PDF. I am thinking about extending the CareKit for such pdf generating VCs but it would take some time.

umerkhan-apple:
@bishalg I think it could be beneficial to provide a richer pdf generator using care card buttons along with other carekit private classes. A PR targeting that entirely would be nice. For now closing this one, so we can focus on creating the bigger PR that will include the changes from this one as well.

Issue #54, opened 2016-05-24
CocoaPod ignores project umbrella header, header visibility.


auibrian:
The current podspec causes CocoaPods to rewrite the umbrella header for the framework that it creates so that it is out of sync with the actual umbrella header `CareKit.h`. Various headers are included in the umbrella that probably shouldn't be, and are listed in the CareKit target as 'public' when they should be 'project'.

malhal:
The cocoa pods generated header can probably just be ignored externally. Firstly because it doesn't generate a correct framework header by using #import "" instead of #import <> and also CareKit already has its own header.
https://github.com/CocoaPods/CocoaPods/issues/4482

Pariecemckinney-apple:
Closing due to inactivity. Please feel free to reopen if you would like the conversation to continue.

PR #55, opened 2016-05-24
Fix for #54: updates visibility of header files, adds module map

auibrian:
Addresses https://github.com/carekit-apple/CareKit/issues/54

Update the podspec such that the public/private/project attributes of the various header files match the actual CareKit.xcodeproj.

Add a module map that refers to CareKit.h as the true umbrella header for the framework so that any modifications to that header are reflected in the CocoaPod as well.


rajitss:
Thanks for bringing this up @auibrian. You've got a few valid points here.   
However I would like propose we take a different approach to solve this.  Here's what I suggest:

1. Move the CareKit.modulemap into the CareKit xcproject itself.  That ways it won't need to be referenced in the podspec explicitly and anyone using CareKit directly (outside of CocoaPods) will also benefit from it.
1. For the podspec `public_header_files` property we might be better off using the same mechanism/script used for ResearchKit. i.e.:   
 `s.public_header_files = './scripts/find_headers.rb  --public'.split("\n")`   
Here's the [link to find_headers.rb](https://github.com/ResearchKit/ResearchKit/blob/master/scripts/find_headers.rb) for reference. This script will help ensure the headers are kept in sync with what's defined in the xcproject itself.
1. Keep the podspec `private_header_files` property as is in stable branch.  The generic pattern will help ensure any private/internal files in the future won't need to be explicitly mentioned. Alternatively this could also be updated to use the `find_headers.rb` script mentioned above, but with the `--private` flag.

(FWIW I think 1 and 2 should probably have been done by us in the first place.)

Thoughts?

auibrian:
Thanks @rajitss! I like the sound of that, some notes:

1. Having the modulemap in the .xcproject is fine for anything (say Carthage) that uses the project to build, but for CocoaPods that's not going to work. In truth the only reason the module map exists is because CocoaPods has no way to specify an umbrella header. We can move the modulemap into the project but we will still need to explicitly reference it in the podspec.

2. Absolutely the `find_headers.rb` script is the way to go. I forgot ReasearchKit had that script, it's a beautiful solution for keeping a pod in sync with the canonical .xcproject.

3. The current pattern currently catches files and marks them as private rather than project, I think the right thing to do here is user `find_headers.rb` with the `--private` switch.

I'm happy to make those changes in my PR once we're in agreement?

rajitss:
Few comments:

1. From what I've been able to gather CocoaPods will automatically create a modulemap if one isn't specified in the podspec -- see the podspec docs [here](https://guides.cocoapods.org/syntax/podspec.html#module_map).  I tried it out and it does infact create one (during a `pod install` or `pod update` phase) if it isn't specified in the podspec, and the user of your pod has the `use_frameworks!` statement in their Podfile. Incidentally the umbrella header statement in the modulemap created is along the lines of  `umbrella header "CareKit-umbrella.h"` where CareKit-umbrella.h includes all the headers mentioned in the podspec's public_header section.
But that said, the modulemap generated is slightly different than the one included in the xcproject. Accordingly, I believe we are better of using your recommendation of explicitly specifying the modulemap in the podspec to ensure consistency between the CocoaPods and non-CocoaPods generated framework.  
I still have one suggestion though, which is to update the modulemap to explicitly include a private module section, similar to what ResearchKit has, as shown [here](https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/module.modulemap).
1. Agreed.
1. Agreed - lets use find_headers.rb script here as well then.

And thanks for offering to make the changes -- please do so if you are ok with the suggested change in 1.

rajitss:
Hi @auibrian, Just checking in to see if you have any comments on the response above. If you are unable to make the changes do let me know as I could help with that.

auibrian:
Hi @rajitss, I'll have the changes ready by tomorrow. WWDC distracted me somewhat :-)

auibrian:
Alright, this should be ready for review now.

rajitss:
Awesome. Thanks! I'll work on getting this merged into master and pushed to Cocoapods.

rajitss:
One comments...Can you please move the CareKit.modulemap into the CareKit project folder? I'm still seeing it in the the root of the repo with the current PR. We would like to keep it inside the CareKit project so that its consistent when CareKit is built without Cocoapods.

auibrian:
@rajitss Taken care of. I also explicitly added the modulemap to the xcode project to maintain consistency. I'm not aware of Xcode automatically picking up module maps on it's own so I figured that was necessary?

rajitss:
Thanks @auibrian. And you are correct about needing to specify the modulemap explicitly in the Xcode project.

rajitss:
Changes look good.

rajitss:
@YuanZhu-apple has pushed the changes to Cocoapods, so these changes should now be live.  Thanks @auibrian for catching this issue and fixing it.

Issue #71, opened 2016-06-17
allow activities to be discontinued


dukemike:
If an app allows the user to add `OCKCarePlanActivity`, it would be useful to allow the user to discontinue them. The Card View could allow a swipe-left gesture on an activity to allow the user to discontinue the activity immediately, or "tomorrow".

Some thoughts on implementing this:
- we would likely not want activities obtained from an EHR to be deferred by the user. `OCKCarePlanActivity` could have a _canBeDeferred_ flag, which would default to _false_.
- the `OCKCareCardViewControllerDelegate` could be expanded to include a `careCardViewController:canDeferInterventionActivity:` operation which would allow developers to specify whether or not an activity can be deferred, and could be called as `OCKCareCardTableViewCell`s are being created in `OCKCareCareViewController`'s `tableView:cellForRowAtIndexPath`.

@umerkhan-apple: We discussed leveraging `careCardViewController:willDisplayEvents...`, but I don't see how that would be used after looking at the code. I believe we were thinking that the developer would specify which events would be deferrable here.

`OCKCarePlanEvent`s would need similar functionality to allow them be discontinued as well.

YuanZhu-apple:
@dukemike Just want to confirm if `discontinue` means `end` an activity? If so, that's straight forward, just set the end date on the schedule of the activity. 

But deferring an event or events could be much more complex, because it involves the scheduling.
For example,  an weekly repeated schedule, should it be deferred to next day or next week?
And how to track the deferring in the care plan store and so on.

YuanZhu-apple:
Another idea that in my head is to support swipe action menus in the table view of care card or tracker. Basically,  the menu option can be provided by developer through delegate, and the action is up to the developer as well.

dukemike:
Yes, I am proposing that swiping on the row in the card card would cause the activity's end date to be set to _now_ or to tomorrow at 0:00. Deferring an event would have many more edges which I could see as rescheduling. 

Do you think that allowing the developer to specify the menu would be better than adding a '...canDefer...' to supporting ending the activity only?

umerkhan-apple:
@dukemike It was great meeting with you at WWDC. I think this will be a great addition to the Care Card.

We probably don't want to add an extra flag to the activity model. 
This will create data migration issues.

Here's how I think we should proceed for this feature:
1. We should instead add a method to the existing **Care Card Delegate**, something like:
[...]

PR #80, opened 2016-07-01
allow activities to be discontinued. refs #71

dukemike:


scdi:
Hi:
How does an app allow users to add OCKCarePlanActivity? Is there some code sample available?

dukemike:
@scdi `OCKCareCardViewController` does not currently support adding an activity, but I know that the folks at Apple have discussed it. I'm working on something that will likely need it and will definitely submit it if I do. That said, if you're looking for a hack, you could add another bar button item to the care card view controller. That would allow you to present your own view controller to create an `OCKCarePlanActivity`.

CareKit/CareCard/OCKCareCardViewController.m
```diff
@@ -474,6 +474,20 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
 }
 
 
+- (void)tableView:(UITableView *)tableView commitEditingStyle:(UITableViewCellEditingStyle)editingStyle forRowAtIndexPath:(NSIndexPath *)indexPath {
+    OCKCarePlanActivity *selectedActivity = _events[indexPath.row].firstObject.activity;
+    OCKCarePlanEvent *event = _events[indexPath.row].firstObject;
```
umerkhan-apple: Refactor to:
```
OCKCarePlanEvent *event = _events[indexPath.row].firstObject;
OCKCarePlanActivity *activity = event.activity;
```

CareKit/CareCard/OCKCareCardViewController.m
```diff
@@ -474,6 +474,20 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
+
+    [self.store setEndDate:event.date
+               forActivity:selectedActivity
+                completion:^(BOOL success, OCKCarePlanActivity * _Nullable activity, NSError * _Nullable error) {
+                }];
```
umerkhan-apple: Use completion block to make sure the end date was set successfully. If not, maybe log it, or show an alert to the user?

CareKit/Localization/en.lproj/CareKit.strings
```diff
@@ -32,6 +32,7 @@
 /* Care Card */
 "CARE_CARD_HEADER_TITLE" = "Care Completion";
 "TODAY_BUTTON_TITLE" = "Today";
+"CARD_CARD_DELETE_TITLE" = "Discontinue After Today";
```
umerkhan-apple: Rather long title for a button. Not sure what a better one would be though.

umerkhan-apple:
@dukemike Minor nits. Overall looks great! 👍 
@YuanZhu-apple Could you also take a look? :)

CareKit/CareCard/OCKCareCardViewController.m
dukemike: @umerkhan-apple Are there any other cases when CareKit is showing a notification that something didn't get saved or changed? I thought about it but didn't think throwing an exception was the right thing to do. Would adding `careCardViewController:didDiscontinueActivity:error` be a better way?

dukemike:
@umerkhan-apple, @YuanZhu-apple Please review. Thanks!

CareKit/CareCard/OCKCareCardViewController.h
```diff
@@ -57,6 +57,18 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
+ Asks the delegate to verify that the given activity can be discontinued [deleted]. If this method is not implemented, all
+ all rows are assumed to not be discontinuable.
+ 
+ If returned YES, and the user chooses to discontinue the activity, the activity's end date will be set to the date of the
+ event that was selected, thereby removing all subsequent events.
+ 
+ @param viewController              The view controller providing the callback.
+ @param interventionActivity        The intervention activity that the user selected.
+ */
```
umerkhan-apple: Update documentation to state that the discontinue will happen after "today's" date. And this is only applicable to the current day. If the user goes back to a previous day, they cannot mark it to be discontinued.

umerkhan-apple:
@dukemike Looks great. Thanks for the updating the test app as well. One thing that is still not fixed is that the user can go back to a previous day, and swipe left on an activity to discontinue it.

The discontinue button should only be available for activities on today's date.

dukemike:
Being able to discontinue an activity on a prior day is something that I'd like to allow. Could preventing discontinuation in the past be enforced by the developer vs the framework? I can imagine the case where the user stopped performing an activity yesterday but didn't open the app to discontinue it then.

umerkhan-apple:
@dukemike Ah that use-case definitely make sense. My concern is that there is no way to allow the developer to discontinue in the past or not. The delegate callback provides the Activity.

One bad path is:
Let's say today is Friday. The user completes Outdoor Walk. Then goes back to Wednesday and discontinues Outdoor Walk. That seems like an error-prone state.

Any creative ideas on how we can go around this?

umerkhan-apple:
Also, one other thing I thought of. Could we just have the button say `Discontinue`, and when the user taps it, just set the endDate to be yesterday?

dukemike:
@umerkhan-apple The idea of checking for future completed events for an activity to determine if the discontinue should be shown sounds both problematic and non-performant. How about if a user tries to discontinue and there are future completed events then: the activity's completion date could be set to after the last completed event, or the user could be asked if they want to delete the future ones? I lean towards asking the user.

re: the discontinuation text / date, my preference is that discontinuation is something that end the activity and does not remove it from view because it feels less like Delete to me, but I'm not tied to it and don't have any major concerns if we set the endDate to yesterday. "Delete" and "Discontinue after today" _feel_ clear to me, while "Discontinue" _feels_ less so, but I agree that it is really wordy.

tktsubota:
I played with this a little bit, and I think it's a great addition (next up is adding activities!) However, I do think that in addition to having it available on swipe, it should be available on the default detail view controller of interventions (assuming the delegate method is implemented).

As for the naming, "Discontinue After Today" is quite long (more than half the width on iPhone 6s). I was thinking maybe "End Today" or "Discontinue" like what @umerkhan-apple said.

As for the error case, there are one of two ways we could go about this:

1. No matter which day is being viewed, discontinuing it would set the end date to today. This seems slightly unintuitive, but changing the title to "Discontinue Today" or "End Today" _might_ work.
2. Discontinue on the day being viewed, and gracefully handle the error if necessary (I'm really not sure how often this particular case will occur). I'm thinking an alert saying "Data after \*_date being viewed_\* for this activity will be lost". There, "OK" or "Cancel" buttons can be provided.

umerkhan-apple:
@dukemike I think for the wording, it would make sense to just say `Discontinue` and have the functionality be that it sets the endDate to be the previous date (or whatever the last date for that activity was). This use-case makes more sense to me, because mostly users will not know ahead of time if they are discontinue for tomorrow or not. For example, when they see a med on their Care Card today, they will want to hit discontinue because they know they are no longer taking it. They won't really hit discontinue because they know they won't be taking it starting tomorrow.

This really gets tricky if it is something that is only once a month. So on the first of every month, they go for physical therapy. Then when they complete that task, they know they will no longer be repeating it the following month, but most likely users won't remember to mark it as discontinue. Cause they will have already logged it as something that they have done. Come next month, they see the activity again. Now if they mark it as discontinued, it won't show up the following month, but it still will for the current month. At this point, they will have to go back to the previous month's activity, and mark that as discontinued. 

This approach seems very counter-intuivite. I think we should just have the button say `Discontinue` and mark whenever the last time the activity occurred as the end date.


For the case where they are going back in time to discontinue something, we can show them an alert confirming that they will lose all the data from that day onwards. If they confirm, then we set the end date and make sure that no data is stored past that end date for the activity.

scdi:
That seems very reasonable to me. That is a good approach.

dukemike:
Sounds good. I'll set the endDate to the previous date and make the other changes.

srinathtm-apple:
Closing this PR as it is no longer relevant due to CareKit 2.0 beta update, however we are discussing internally to provide support for discontinuing tasks...

PR #85, opened 2016-07-07
OCKChartLegendView has more than one row with NSLayoutRelationLessThanOrEqual in NSLayoutAttributeWidth inside UITableViewCell causes layoutSubviews to be called when scrolling only

weijentu:
Owing to the issue only happens when `_legendView` has more than one row, please modify `OCKBarChart` little bit to begin with.

First, please change `[painBarSeries, medicationBarSeries]` to `[painBarSeries, medicationBarSeries, painBarSeries, medicationBarSeries]` in the class `BuildInsightsOperation.swift`

Than `OCKInsightsViewController` will looks like the following screenshot:
simulator screen shot jul 7 2016 2 47 15 pm

And now, let's set a break point in the `- (void)layoutSubviews` of the class `OCKGroupedBarChartView`. Then, we will see it's called when we just scroll the table only.

Change `NSLayoutRelationLessThanOrEqual` to `NSLayoutRelationEqual` can omit above issue.


YuanZhu-apple:
I don't see any wrong with the screenshot you attached.
@weijentu Can you attach the screenshot after this change as well?

weijentu:
Hi @YuanZhu-apple , please take a look of bottom part of the screenshot. There are two rows for `_legendView' and "[colorBox] Pain" and "[colorBox] Medication Adherence" are duplicated. At a state like this, It adversely causes scrolling lag because `layoutSubviews` is called when scrolling.

YuanZhu-apple:
what's wrong with 4 legends for 4 bar series?

weijentu:
The reason to have 4 legends for 4 bar series is to produce this performance issue. Actually, any modification has `_lengendView` with more than one row can causes it.

For example:
In the 2 legends for 2 bar series, `layoutSubviews` will not be called when table is scrolling.
In the 4 legends for 4 bar series, `layoutSubviews` will be called when table is scrolling.

Set a break point in the `- (void)layoutSubviews` of the class `OCKGroupedBarChartView` can see the difference.
When a table is scrolling and accompanying with `layoutSubviews` keeps being called, scrolling  is quite lag in this scenario.

weijentu:
Please forgive me that English is not my native language and I might not explain it very well. I just recorded a video to elaborate what I mean. In the video, please notice on the situation that `- (void)layoutSubviews` keeps being called when I was trying to scroll the table.

[https://db.tt/0jW9UaHx](url)

YuanZhu-apple:
@weijentu I only see it get called once, when there are 4 legends. 

screen shot 2016-07-07 at 4 20 18 pm

YuanZhu-apple:
@weijentu I get it, thanks!

YuanZhu-apple:
r=me.

PR #86, opened 2016-07-08
Added methods to OCKCareCardViewControllerDelegate - provide customization

pedrohugorm:
Added methods to customize TableHeader, TableFooter and WeekViewController using delegates.


umerkhan-apple:
@pedrohugorm Thank you for this great addition. I have looked at the PR thoroughly, but will get to it soon. However in my initial skim, I did notice that the Test App has not been updated. Could you please add an example use-case for this into ORKTest?

umerkhan-apple:
@pedrohugorm Any updates on this one?
Could you also remove the additional strings file?

Issue #108, opened 2016-08-23
CareKit in Storyboards


tktsubota:
See #109 for the respective PR.

Bringing OCKCareCardViewController, OCKSymptomTrackerViewController, OCKInsightsViewController, and OCKContactsViewController to storyboards is a highly requested feature (see #68), and I would also like to use these in storyboards.

In order for this to be possible for OCKCareCardViewController and OCKSymptomTrackerViewController, an OCKCarePlanStore singleton needs to be provided. For OCKInsightsViewController and OCKContactsViewController, only an `initWithCoder:` implementation needs to be added.
## OCKCarePlanStore Singleton

Because of how storyboards work, you cannot supply a care plan store on initialization. While a possible alternative is subclassing and overriding `initWithCoder:`, you cannot set the `store` because it is defined as read-only (this might be possible in Objective-C, but you definitely cannot in Swift without terrible hacks).

[CareKit apps only need one care plan store](http://carekit.org/docs/docs/CreatingTheCareCard/CreatingTheCareCard.html). Therefore, the developer shouldn't have to create one and manage it themselves, making sure they pass it to every class that needs it, and not creating duplicates.

Therefore, a singleton care plan store can both make CareKit a ["faster 'get started' to a useful app"](https://github.com/carekit-apple/CareKit/blob/master/CONTRIBUTING.md) and a way to get CareKit in storyboards.
### Implementation
1. **[Add a singleton](http://www.galloway.me.uk/tutorials/singleton-classes/) to `OCKCarePlanStore`**. It should be called something simple like `defaultStore` and have a default URL for storing the data (probably in the Document Directory in the User Domain Mask like in the [programming guide](http://carekit.org/docs/docs/CreatingTheCareCard/CreatingTheCareCard.html#InstantiatingTheCarePlanStore)).
2. **Make the `store` parameter from the view controllers' initializers `nullable`, using the singleton store if passed `nil`.**
3. **Set `store` in `initWithCoder:` to the singleton.** Set `store` in `initWithCarePlanStore:` to the singleton, unless the `store` parameter is not `nil`.

You will be able to access the app's care plan store via `OCKCarePlanStore.defaultStore()` or `[OCKCarePlanStore defaultStore]`.
### Impact on Existing Code

Making the `store` parameter `nullable` will keep existing code working while providing an opportunity to remove all care plan store management and being able to use storyboards.

Internal CareKit code won't need a huge change. The only change will be in the initializers of the view controllers.

[...]

PR #109, opened 2016-08-23
Add Storyboard Support

tktsubota:
**Updated 12/31/16: See Comment Below**

---

This adds support for issue #108.
### The Two Features

There are two major features in this: the default care plan store (`OCKCarePlanStore.defaultStore()`) and `initWithCoder:` implementations to support storyboard usage.
### Implementation

I essentially followed what I outlined in #108.
#### Actual Source

`OCKCarePlanStore.defaultStore()` is the big one here. I added it as a class method, internally calling `initWithPersistenceDirectoryURL:` with a directory named "CarePlanStore" in the user's document directory.

`initWithCoder:` in the view controllers essentially duplicates the existing initializers, without the parameter passed in.

Most of the public properties of the view controllers were made `IBInspectable` so we can modify those properties in interface builder without needing to subclass.

`initWithPersistenceDirectoryURL:` in OCKCarePlanStore and `initWithCarePlanStore:` in the care card and symptom tracker were left to be, so these changes wouldn't be source-breaking and we can still use a custom care plan store if we want. `init` of the view controllers was made available, using the default store.
#### Documentation

With these two major features came with a huge change in the programming guides. I switched to using the default store for everything, while still providing an option to create a custom store if need be. The view controller creation in the storyboard was centralized in "Creating The Care Card".

I also had to revert #14's changes to the code-blocks, because when I ran the "CareKit-docs" build target to generate the appledoc, I found out appledoc didn't support the fenced code-blocks and it didn't look good at all. I also fixed broken links.
#### Test Apps

I wasn't sure if I should have used the singleton throughout the entire OCKTest (and even OCKSample), so I left it be and made a separate "Storyboard" test in OCKTest, where every view controller there is instantiated in the storyboard and uses the default store.
#### Unit Test

I also added a unit test that duplicates some of the existing care plan store test, just to make sure the default care plan store was working.
### Bugs I found
- When changing the tint color on the care card, the heart image color does not change. I know of a fix, but I didn't put it in this PR as it didn't seem to relate.
### Future Improvements
- We should start moving toward using the default store in the documentation, test apps, and so on. This should be the recommended way of using the care plan store, as there aren't a lot of reasons to use a custom URL.
- We should consider switching from using appledoc to [Jazzy](https://github.com/realm/jazzy). While they just recently got support for Objective-C, it looks more like the current documentation (although the look of documentation changed with iOS 10) and has a larger community. I noticed a lot of bugs and missing features in appledoc that I had to work around. 

I expect a lot of changes from code review, as this is a major feature (and _lots_ of changes in the programming guides).


tktsubota:
Updates to finish off the year...

### Xcode 8 Class Property

Xcode 8 came with the ability to create [class properties](http://blog.andrewmadsen.com/post/145919242155/objective-c-class-properties) in Objective-C to better interoperate with Swift. I converted the default store to a class property named `defaultStore`. It can be accessed by:
- Objective-C dot notation: `OCKCarePlanStore.defaultStore`
- Objective-C bracket notation: `[OCKCarePlanStore defaultStore]`
- Swift: `OCKCarePlanStore.default` (NOT `defaultStore`)

Everything else related to the singleton remained the same.

### Swift 3 OCKTest Update

The storyboard files were updated to Swift 3 and the default store property.

I also noticed that there are quite a few crash bugs in OCKTest that should be fixed.

### Removed Documentation Changes

To narrow the scope of this PR, I removed the changes in documentation. They can be put back in (with the power of git) if need be.

tktsubota:
@umerkhan-apple Any updates on this PR?

umerkhan-apple:
@tktsubota Putting this PR on hold for now. Will test internally with the rest of the team and get back to you.

tktsubota:
Reopening later.

PR #122, opened 2016-10-06
Update the ResearchKit dependency to the 1.4 Convergence commit

programmarchy:
The ResearchKit submodule was pointed to an older commit, which prevented CareKit from compiling.

This PR updates the submodule to the "Convergence" commit for ResearchKit.


umerkhan-apple:
Good catch! Thanks! r=me.

PR #130, opened 2016-12-03
Public access to table view properties

moppymopperson:
I'm working on a project in which I'd like to add a custom `tableViewHeader` to some of the table views in CareKit, but there is no public accessors currently.

The workaround I'm currently using is to filter subviews for a `UITableView`, but this is not really an ideal solution, so I decided to create a pull request to make the tableView property on `OCKCareCardViewController`, `OCKConnectViewController`, and `OCKSymptomTrackerViewController` available publicly.

### DIFF
--- a/CareKit/CareCard/OCKCareCardViewController.h
+++ b/CareKit/CareCard/OCKCareCardViewController.h
@@ -144,6 +144,11 @@ OCK_CLASS_AVAILABLE
  */
 @property (nonatomic, readonly, nullable) OCKCarePlanEvent *lastSelectedInterventionEvent;
 
+/** 
+ A reference to the `UITableView` contained in the view controller
+ */
+@property (nonatomic, nonnull) UITableView *tableView;
+
 /**
  The image that will be used to mask the fill shape in the header view.
  
--- a/CareKit/CareCard/OCKCareCardViewController.h
+++ b/CareKit/CareCard/OCKCareCardViewController.h
@@ -1,5 +1,6 @@
 /*
  Copyright (c) 2016, Apple Inc. All rights reserved.
+ Copyright (c) 2016, Erik Hornberger. All rights reserved.
  
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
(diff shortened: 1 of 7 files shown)


CareKit/CareCard/OCKCareCardViewController.h
```diff
@@ -144,6 +145,11 @@ OCK_CLASS_AVAILABLE
  */
 @property (nonatomic, readonly, nullable) OCKCarePlanEvent *lastSelectedInterventionEvent;
 
+/** 
+ A reference to the `UITableView` contained in the view controller
+ */
+@property (nonatomic, nonnull) UITableView *tableView;
```
umerkhan-apple: Should we make this readonly?

moppymopperson: I had it set to allow writing as well so that it would be possible to replace the default `tableViewHeader` with a different. I believe setting the tableview to readonly would prevent setting it's `tableViewHeader` property, wouldn't? 

If I'm mistaken about that, then readonly would be appropriate.

Another possible compromise between the two would be to add public getters and setters that jump over the tableView and give direct access to the table views' headers.

umerkhan-apple: When you make the tableView readonly, it does not apply to any of its properties. The properties have their own read/write specifier. As long as the tableViewHeader is not defined as readonly (which it is not), setting the tableView as readonly should still allow you to set the tableViewHeader.

moppymopperson: In that case, let's go with your suggestion to make it private.

umerkhan-apple: @moppymopperson Great, could you update the PR?

umerkhan-apple: @moppymopperson Great, could you update the PR?

moppymopperson:
Gladly. I just pushed the readonly change request.

umerkhan-apple:
@moppymopperson Thanks for updating the PR! Looks good! 👍

PR #160, opened 2017-04-30
Care card button optional image

moppymopperson:
This PR adds a `buttonImage` property to `OCKCareCardViewController`. Setting this optional property populates an image view on the contained `OCKCareCardButton`s. The image is not displayed when the buttons are unchecked, or if the `buttonImage` property is left nil. 

Under the hood, setting `buttonImage` passes the image to a similarly named property in all instances of `OCKCareCardTableViewCell`, which in turn passes it down to all instances of `OCKCareCardButton`, where it is added to an image view.


moppymopperson:
Updated and moved to #233

PR #169, opened 2017-06-09
Synchronize Lazy Instantiation of Schedule Calendar

apbendi:
The `UTC_calendar` property on `OCKCareSchedule` is initialized
and configured lazily on first access inside a getter. This operation
is not atomic. Because the framework may call schedule methods on
different threads (for example, some operations are on
the OCKCarePlan serial), this can create a race condition.

To resolve this issue, we synchronizing the lazy initialization on
the instance itself. This resolved crashes we'd been seeing related
to this property.

### DIFF
--- a/CareKit/CarePlan/OCKCareSchedule.m
+++ b/CareKit/CarePlan/OCKCareSchedule.m
@@ -199,9 +199,11 @@
 }
 
 - (NSCalendar *)UTC_calendar {
-    if (!_calendar) {
-        _calendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
-        _calendar.timeZone = [NSTimeZone timeZoneWithAbbreviation:@"UTC"];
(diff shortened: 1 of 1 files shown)


kelseydedoshka-apple:
@apbendi - looks good, thanks for contributing this back!
@akshay-yadav-apple - please take a look!

UberJason:
@apbendi I'm having occasional crashes in NSDateComponents+CarePlan.m, and I believe the same it may be the same issue - do you agree? See [here](https://github.com/carekit-apple/CareKit/blob/e9347ecba9635c7889fcdf37ca64afd629960322/CareKit/CarePlan/NSDateComponents%2BCarePlan.m#L120).

apbendi:
@UberJason, the pattern here looks very similar to the one where I saw the issue, yes. Try running with the thread sanitizer turned on. That should confirm if the crashes are due to a thread contention issue.

PR #171, opened 2017-06-10
Add visualGroupIdentifier to OCKCarePlanActivity

### LINKED ISSUE #170: OCKCareContentsViewController uses groupIdentifier to determine tableView sections - unclear documentation
I noticed that `OCKCareContentsViewController` is using the `groupIdentifier` of `OCKCarePlanActivity` to determine visual groupings of different activities in the Care Contents view (tableView sections). However, the documentation for `groupIdentifier` implies that its purpose is more for internal organization and querying purposes:

> Group identifier string. You can use the identifier to group similar activities. You can use this identifier as a key to retrieve multiple activity instances from the store.

And indeed, `OCKCarePlanStore` has a method `activitiesWithGroupIdentifier:completion:`.

Furthermore, if an activity is `optional`, then its `groupIdentifier` is ignored for visual organization, and it is instead placed in the "Optional" tableView section. So the usage of `groupIdentifier` in determining visual groupings is somewhat inconsistent.

I think there would be value in having a different property on `OCKCarePlanActivity` for visual groupings of activities in the Care Contents view that doesn't clash with `groupIdentifier`. For example, I could imagine wanting to group activities by the provider who assigned them using `groupIdentifier`, while still visually separating the activities in the Care Contents view by, for example, their activity types (e.g. Physical Exercises vs. Medications).

UberJason:
Resolves #170 by adding a new property on OCKCarePlanActivity, `visualGroupIdentifier`. Activities grouped with the same `visualGroupIdentifier` will appear together in the Care Contents card. The documentation note also explicitly mentions that this behavior is overridden if the activity is `optional` or `read-only`. This preserves the `groupIdentifier` field to group activities for querying in the care plan store, as its original intention was.

Specific changes in this PR:
* Added the `visualGroupingIdentifier` property on OCKCarePlanActivity.
* Updated the backing Core Data model with a new model version (v3) that includes the `visualGroupIdentifier`. 
* Updated OCKCareContentsViewController to use `visualGroupIdentifier` instead of `groupIdentifier` to form the tableView sections. 
* Updated sample project to use `visualGroupIdentifier`, showing two different visual groupings - Assessments and Physical Activity.
* Updated tests to use initializer with `visualGroupIdentifier`.


UberJason:
@kelseydedoshka-apple Did you ever have a chance to take a look at this PR since our discussion around issue #170? I still can't work on any of the documentation until the 1.2 spec with updated documentation comes out, but was hoping you could take a look at the current state to discuss if any other changes might be needed.

kelseydedoshka-apple:
@UberJason - major apologies on the long delay here. Would you mind resolving the merge conflicts on this?

UberJason:
@kelseydedoshka-apple I've updated to fix the merge conflicts. 

I've also made a very minor update to the sample project that shows an example of a divide between the `groupIdentifier` and `visualGroupIdentifier`: in this example, all Assessment type activities have a group identifier "Assessments" (so they can be queried together), but the assessments are divided into two visual groupings - "0-10 Scale Assessments" vs. "Measurement Assessments". (It's admittedly a little contrived, but illustrates the intent of this enhancement.)

CareKit.xcodeproj/project.pbxproj
```diff
@@ -1482,10 +1483,11 @@
 		8677EE041C9775EB00588CD6 /* OCKCarePlanStore.xcdatamodeld */ = {
 			isa = XCVersionGroup;
 			children = (
+				96AED2731EEC933400C98B5B /* OCKCarePlanStore v 3.xcdatamodel */,
 				185BC3A31D81DF0F001AB0A9 /* OCKCarePlanStore v2.xcdatamodel */,
 				8677EE051C9775EB00588CD6 /* OCKCarePlanStore.xcdatamodel */,
 			);
-			currentVersion = 185BC3A31D81DF0F001AB0A9 /* OCKCarePlanStore v2.xcdatamodel */;
+			currentVersion = 96AED2731EEC933400C98B5B /* OCKCarePlanStore v 3.xcdatamodel */;
```
kelseydedoshka-apple: Would you mind keeping the same format here on the data models? So OCKCarePlanStore v3 without the space between v and 3?

kelseydedoshka-apple:
@UberJason - overall this is looking really good and I appreciate you adding some examples to the sample app.

Just for my understanding as well, the group identifier in this case still operates the same as it always has, and can be used to query the care plan store on? The visual Identifier, indicates how the items are group in the view. If the user does not set a visual identifier, but has a group identifier, does it default to grouping items visually by group identifier or no?

Also as this is a bigger change, would like to get a second reviewer to look at this as well. @akshay-yadav-apple - do you mind performing a review on this?

CareKit.xcodeproj/project.pbxproj
UberJason: @kelseydedoshka-apple Of course! I've pushed that update. (Not sure how that space got in there, I was pretty sure I just used Xcode to add a new model version, but oh well!)

UberJason:
@kelseydedoshka-apple Correct - the `groupIdentifier` operates the way it always has, and now has the sole purpose of being used to query the care plan store. The `visualGroupIdentifier` has the sole purpose of grouping items in the view.

No, if there's no `visualGroupIdentifier`, it does not fall back to grouping by `groupIdentifier`. From a philosophical standpoint, I was trying to keep a more clean separation of grouping-for-querying (`groupIdentifier`) and grouping-for-display (`visualGroupIdentifier`); also, practically speaking, I didn't want to add even more complexity by falling back on groupIdentifier for display (what if some items have a groupIdentifier and a visualGroupIdentifier, while others have only one or the other?). If an item has no `visualGroupIdentifier`, it gets grouped into the "Other" group.

One thing I hadn't considered until this very moment was schema migration from v2 to v3 - if someone gets an app update with v3 of the care plan store, all existing items will get grouped into "Other", because now CareKit will be looking for `visualGroupIdentifiers` (which didn't exist in v2 when those items were created). Should I add a Core Data data migration that automatically copies `groupIdentifier` to `visualGroupIdentifier`?

kelseydedoshka-apple:
@UberJason - Definitely understand your thoughts on wanting to keep the group and visual group identifiers separate. Thats a really great point, I think it would be good to have that migration.

kelseydedoshka-apple:
@UberJason - great thanks!

akshay-yadav-apple:
Hey @UberJason 
 Thanks for the PR. This PR handles a very important issue.
Since the PR introduces a new parameter in the initializer, the existing apps will crash if the framework is updated. We strongly want to avoid this, and make the changes backward compatible for the existing apps. Also, since the group identifier will no longer be used as section headers, the CareContents page will show a list of all the activities altogether, without any section headers. We would love to hear your thoughts on how we can avoid these issues, and achieve the functionality.

UberJason:
@akshay-yadav-apple I believe my latest commit adding the Core Data mapping model addresses both of your concerns. That mapping model copies `groupIdentifier` to `visualGroupIdentifier`, preserving the existing grouping for apps that were running v2 of the model. Because the Core Data stack in CareKit already uses automatic version migration, the app won't crash upon update - it'll migrate to v3 and use the existing grouping quietly.

The way I tested to ensure this behavior was the following, and feel free to try it yourself (and let me know if it doesn't work!):
1. Check out the latest build from `master` and run it fresh, creating the app with the v2 Core Data model.
2. Check out my PR branch and run it directly on the same phone, simulating an App Store update. 
3. Visually verify that the groupings are identical, and use the debugger to check that the `visualGroupIdentifier` property for any given activity is now populated with the value from `groupIdentifier`. (This verifies that the model was updated from v2 to v3.)

UberJason:
Oh, sorry - I just noticed you were referencing the initializer, not just the model. Let me think on an appropriate fix, but my immediate thought is that we could deprecate the old initializer but keep it working, and have it under-the-hood call the new initializer, setting passing the `groupIdentifier` into the `visualGroupIdentifier` parameter.

UberJason:
@akshay-yadav-apple I've pushed two more commits to this PR:
1. I've added back the old initializers that don't reference `visualGroupIdentifier` and marked them as deprecated, and implemented them as directly calling their newer initializer counterparts that use `visualGroupIdentifier`, and using the value of `groupIdentifier` for `visualGroupIdentifier` as well.
2. I noticed that I forgot to add my Core Data mapping model to the list of Resources in the Podspec - so when CareKit was updated as a Cocoapod, the mapping model wasn't included, and all the old groupings were lost (all activities were shown together without section headers). Was this the behavior you were referencing yesterday? I've fixed that issue and now verified that updating CareKit via Cocoapod will include the mapping model, and migrates the Core Data model version smoothly.

akshay-yadav-apple:
@UberJason The recent commits address both of my concerns. Since the old initializers (now deprecated) still exists, and assign value of 'groupIdentifier' to 'visualGroupIdentifier', the existing apps should behave as they were. 
Thank you for making the requested changes. I'll do a final review, before merging the PR.

UberJason:
@akshay-yadav-apple Hey there! Just wondering if there's any update on this? Have you had a chance to do a final review, is there anything else I can do to help?

srinathtm-apple:
Hey @UberJason , we are still in the process of reviewing this PR. We need to account for backward compatibility checks to make sure that nothing breaks because of this change, thank you for your patience!

UberJason:
Hi @akshay-yadav-apple! I thought I would wait until WWDC this year to see if there were any new updates to CareKit coming that might supersede or include this PR, but I didn't see any in yesterday's session. Therefore, I was wondering if your team still plans to review this PR and potentially merge this enhancement?

akshay-yadav-apple:
Hi @UberJason thanks for reaching out again and apologies for the delay on this!  You are correct that we did not provide any updates in yesterday's session but that does not mean we have forgotten this.  Thanks for your patience and we will be actively reviewing in the coming weeks!

srinathtm-apple:
Closing out this PR as this is no longer relevant due to the CareKit 2.0 beta update..

Issue #177, opened 2017-06-24
Unsatisfiable constraints for Insights View Controller if instantiated with zero widgets


UberJason:
In CareKit 1.2, if `OCKInsightsViewController` is instantiated without any widgets, for example:

`let viewController = OCKInsightsViewController(insightItems: insights, patientWidgets: [], thresholds: nil, store: storeManager.store)`

Then when the OCKInsightsViewController appears on screen for the first time, Auto Layout prints to console warning about simultaneously satisfying constraints:

```
[LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x608000296440 UIStackView:0x7ff4c4f1e8c0.height >= 90   (active)>",
    "<NSLayoutConstraint:0x6080002963a0 V:|-(0)-[UIStackView:0x7ff4c4f1e8c0]   (active, names: '|':OCKInsightsTableViewHeaderView:0x7ff4c4f1d3f0 )>",
    "<NSLayoutConstraint:0x608000295130 UIStackView:0x7ff4c4f1e8c0.bottom == OCKInsightsTableViewHeaderView:0x7ff4c4f1d3f0.bottom   (active)>",
    "<NSLayoutConstraint:0x6080002978e0 OCKInsightsTableViewHeaderView:0x7ff4c4f1d3f0.height == 0   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x608000296440 UIStackView:0x7ff4c4f1e8c0.height >= 90   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKit/UIView.h> may also be helpful.
2017-06-23 23:49:20.333109-0400 Nerve Pain[19504:576028] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
[...]

PR #178, opened 2017-06-24
Fix for issue 177 - OCKInsightsTableViewHeaderView conflicting constraints

UberJason:
Fix for #177 - reduce the priority of OCKInsightsTableViewHeaderView's internal height constraint to 999, so any owner of this header view can set its height with higher priority.


UberJason:
@kelseydedoshka-apple Did you ever have a chance to take a look at this PR since our discussion around issue #170? I still can't work on any of the documentation until the 1.2 spec with updated documentation comes out, but was hoping you could take a look at the current state to discuss if any other changes might be needed.

akshay-yadav-apple:
@UberJason 

Thanks for the PR, it looks good!
Merging it in.

PR #176, opened 2017-06-19
Fix for issue 175: "OCKCarePlanStore - Unwired property (directoryURL)".

mirlhaq:
Linked issue: #175 .

PR #180, opened 2017-06-26
Implement getter method for the directoryURL property on OCKCarePlanStore

apbendi:
Property was simply not wired to it's ivar.


apbendi:
This was fixed with the merger of #176

Issue #182, opened 2017-06-28
headerTitle not working on CK 1.2


ninoguba:
The headerTitle property added to the Care Card and Symptom Tracker views on PR #141  doesnt work anymore in the latest version of CareKit (1.2)

kelseydedoshka-apple:
@ninoguba - thanks for pointing this out. Just to clarify - is this no longer working for the existing Care Card and Symptom Tracker views, or this is feature not accessible via the new Care Contents view?

ninoguba:
@kelseydedoshka-apple not working for the existing Care Card and Symptom Tracker views.

kelseydedoshka-apple:
@ninoguba - thanks for the additional clarification here. Stay tuned for updates on this.
@akshay-yadav-apple , @Pariecemckinney-apple - please take a look at this.

ninoguba:
Submitted PR #189 to resolve this.

Pariecemckinney-apple:
@ninoguba The PR was merged. Thank you for the contribution.

PR #189, opened 2017-07-07
Fix for displaying custom headerTitle when set

ninoguba:
This addresses issue #182.

Starting in CK 1.2 the following localization strings are no longer used:

- CARE_CARD_HEADER_TITLE
- SYMPTOM_TRACKER_HEADER_TITLE

They have now been replaced by the following:

- HEADER_TITLE_CARE_OVERVIEW
- HEADER_TITLE_ACTIVITY_STATUS


Pariecemckinney-apple:
@ninoguba, Looks good on my end

@kelseydedoshka-apple, could you also take a look?

kelseydedoshka-apple:
@Pariecemckinney-apple - Looks good!

PR #191, opened 2017-07-08
Close #190

### LINKED ISSUE #190: Missing () following default
The sample app does not currently build in XCode 8 or 9 beta.

In 8 it does not build because of line 136 in the watch app's InterfaceController.swift file.
`tableView.insertRows(at: IndexSet.init(integersIn: Range(NSMakeRange(rowIndex, activityEventRows)) ?? 0..<0) , withRowType: "EventRow")`. This compiles fine under XCode 9 beta, so I believe it is related to a Swift 3/4 difference.

In 9 it does not build because of the lines  50 and 207 in the WatchConnectivityManager. Both use `WCSession.default` but should be `WCSession.default()`.

moppymopperson:
See #190 
Update to `WCSession.default` to `WCSession.default()`

### DIFF
--- a/Sample/OCKSample/WatchConnectivityManager.swift
+++ b/Sample/OCKSample/WatchConnectivityManager.swift
@@ -47,7 +47,7 @@ class WatchConnectivityManager : NSObject {
     
     init(withStore store : OCKCarePlanStore) {
         self.store = store
-        self.session = WCSession.default
+        self.session = WCSession.default()
 
         super.init()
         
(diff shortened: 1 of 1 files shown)


Pariecemckinney-apple:
@moppymopperson, Thank you for the PR. This fix will be accepted.

Pariecemckinney-apple:
Looks good

@akshay-yadav-apple, could you also take a look

PR #194, opened 2017-07-09
Synchronize lazy instantiation of UTC_gregorianCalendar

UberJason:
Similar to fixed PR #169, this commit synchronizes the lazy instantiation of the UTC_gregorianCalendar to avoid crashes due to race conditions.


kelseydedoshka-apple:
@UberJason looks great - thanks for the fix!
@Pariecemckinney-apple - please take a look!

Issue #185, opened 2017-06-30
Question: We can now swipe forward to future dates on the Care/Symptom views. Is this intentional?


ninoguba:
Is this intentional for CareKit 1.2?

akshay-yadav-apple:
Hey @ninoguba,
Thanks for pointing out this issue, it is not intentional.

UberJason:
I feel like this may actually be a useful feature, though. For example, previewing your upcoming medication/assessment schedule (especially if your care plan is going to change soon, e.g. a provider has prescribed a new medication and ended an old one). Is there a reason to disallow it?

akshay-yadav-apple:
@UberJason,
I agree it would be a good feature to allow viewing the upcoming activities, as long as the user is not allowed to complete any future activity (only view).

ninoguba:
@akshay-yadav-apple so can we expect to have this date checking logic implemented within the framework or is this something you want app developers using CK to implement ourselves in our apps?

ninoguba:
@akshay-yadav-apple submitted PR #199 to fix this unintentional change in CK 1.2.

Pariecemckinney-apple:
@UberJason - We completely agree that this would be a very useful feature for many reasons. However, being that this was a unintentional change, we will accept this merge request and begin discussion on supporting the viewing of future dates. Thanks again for the great input and thank you @ninoguba for the issue and PR .

Closing now

PR #199, opened 2017-07-19
Fix to prevent navigating to future dates in the Care Card, Symptom Tracker, and Care Contents views

ninoguba:
Prior to CK 1.2, navigating to future date in Care Card and Symptom Tracker is prevented.  In CK 1.2 this was broken.  This PR fixes the issue.

### DIFF
--- a/CareKit/CareCard/OCKCareCardViewController.m
+++ b/CareKit/CareCard/OCKCareCardViewController.m
@@ -286,7 +286,8 @@
 }
 
 - (void)setSelectedDate:(NSDateComponents *)selectedDate {
-    _selectedDate = selectedDate;
+    NSDateComponents *today = [self today];
+    _selectedDate = [selectedDate isLaterThan:today] ? today : selectedDate;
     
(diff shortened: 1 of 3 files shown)


UberJason:
Just my opinion here, but as I mentioned in #185, I think it would actually be good to retain the ability for a user to view their future activities, so long as they're unable to interact with them. Disabling swiping to the future entirely, I think, would be a step backward.

ninoguba:
If you review the code prior to my PR. There are actually remnants of code that still does the prevention which tells us that 1.2 was released with a bug actually that only looked like an unintentional enhancement.

In my opinion, it would be better to fix this bug to keep the framework up to its intended spec and let it be the other developers prerogative to implement allowing future date support in their own CK-based apps and contribute it back once properly implemented - for example, to prevent taking results for future activity events.

ninoguba:
The documentation [here](http://carekit.org/docs/docs/Overview/Overview.html) also stated the following:

> They cannot view or edit days in the future.

Pariecemckinney-apple:
@ninoguba - Thank you for the PR. Merging now.

PR #218, opened 2017-10-07
Bug Fix - Enforced group order as how activities get inserted into the care plan when isGrouped=TRUE and isSorted=FALSE

ninoguba:


ninoguba:
@Pariecemckinney-apple please review. thanks!

Pariecemckinney-apple:
@ninoguba Thank you for the PR! Before merging would you mind elaborating on the exact bug that was fixed and how to reproduce it?

ninoguba:
If for example you added the following activities in your Care Plan:

1) Weight (groupIdentifier = "Log")
2) Height (groupIdentifier = "Log")
3) Pain (groupIdentifier = "Survey")
4) Physical (groupIdentifier = "Survey")
5) Wound (groupIdentifier = "Image")

If isGrouped = TRUE and isSorted = FALSE,  you would expect the activities will show in your Care Contents (or other views) grouped in the following order:

1) Log
2) Survey
3) Image

But this is **_NOT_** guaranteed because the underlying data structure used is NSMutableDictionary.  This bug fix preserves the correct "insertion" order by using NSMutableArray instead.

Pariecemckinney-apple:
@ninoguba, Thanks for the explanation and the PR! Merging.

PR #225, opened 2017-12-11
Update connect header when OCKPatient is set

xmlmodeling:
Replacement for previous PR, synced with current CareKit repo and w/o dependency.


kelseydedoshka-apple:
@xmlmodeling - thanks for updating this looks great!

PR #231, opened 2018-01-20
Make table view public on OCKCareCardDetailViewController

moppymopperson:
This PR is to make the tableview in `OCKCareCardDetailViewController` public so that it's header and footer can be customized.

I submitted a handful of PR in the past to make the tableviews in other controllers visible, and I currently have a use case in which I need the same flexibility in the care card detail view controller, so I'm submitting this as an additional PR in the same vein as the previous ones. See #130 for reference.


srinathtm-apple:

@moppymopperson - feel free to re-open this issue if it still exists in CareKit 2.0 ;)

PR #232, opened 2018-01-20
Make OCKConnectDetailViewController public for subclassing

moppymopperson:
This PR is to make `OCKConnectDetailViewController` a public class open to subclassing. We have a use case in which being able to subclass the connect detail view controller is helpful, and having this change adopted into the official repository could help other developers wishing to do something similar.


srinathtm-apple:

PR #233, opened 2018-01-20
Added optional image to the carekit buttons

moppymopperson:
This PR is to add an optional button image for display on the `OCKCareCardButton` shown on the care card and care contents screens. The default is `nil`, so it will not effect current users, but will provide the option to customize the appearance of the buttons to those who would like it.


srinathtm-apple:

PR #242, opened 2018-02-25
More aesthetic arrangement for Care Card buttons

MacroYau:
Currently, each Care Card activity supports up to 14 buttons (intervention events), and the buttons layout is hard-coded to 7 per row, regardless of the screen size. Due to this hard-coded layout, arranging 8 to 12 buttons may look unbalanced.

This PR proposes changing the buttons layout as follows:

1. Maximum number of buttons per row is determined by screen width (e.g., 8 buttons a row is possible on iPhone 8 Plus); and
2. The actual number of buttons per row to be rendered is then calculated with respect to the number of buttons, so as to achieve a more aesthetic layout.

The newly introduced "Drink Water" activity in OCKSample app demonstrates both changes. With the "8 glasses of water a day" interventions, the current CareKit implementation renders the Care Card buttons like this:

> O O O O O O O O
> O

Now, with the screen width-based arrangement, the 8 buttons can be shown in a single row on larger screens. For smaller screens (e.g., iPhone 8 / SE), the buttons will be arranged in a more balanced fashion:

> O O O O
> O O O O

### DIFF
--- a/CareKit/CareCard/OCKCareCardTableViewCell.m
+++ b/CareKit/CareCard/OCKCareCardTableViewCell.m
@@ -1,5 +1,6 @@
 /*
  Copyright (c) 2017, Apple Inc. All rights reserved.
+ Copyright (c) 2018, Macro Yau. All rights reserved.
  
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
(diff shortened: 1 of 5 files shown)


srinathtm-apple:
Hey @MacroYau - closing out this PR due to the CareKit 2.0 beta update, we've added this same behavior by leveraging collection views.

PR #244, opened 2018-02-25
Remove navigation bar tinting from framework

MacroYau:
This PR removes the CareKit view controllers' navigation bar tinting currently hard-coded or altered by `glyphTintColor`. Instead, a navigation bar style consistent throughout the app should be defined in `application(_:didFinishLaunchingWithOptions:)` (as reflected in `AppDelegate` of OCKSample).

For the "Today" button, I believe that it should be colored according to the navigation bar styling, rather than `glyphTintColor`, given the [recommendations in HIG](https://developer.apple.com/ios/human-interface-guidelines/visual-design/color/).


srinathtm-apple:
Closing this PR as it is no longer relevant due to CareKit 2.0 beta update..

PR #247, opened 2018-03-10
support composed character sequences in monograms

fernandomorgan:
This is a simple change to allow support for monograms that include composed character sequences in names


srinathtm-apple:
Closing out this PR as it is no longer relevant due to CareKit 2.0 beta release..

Issue #264, opened 2018-11-18
Project hardening


bgannin:
I've enabled a large number of analyzer and compiler flags to harden the overall project's code including running the analyzer by default and treating warnings as errors. 

CLANG_ANALYZER_SECURITY_FLOATLOOPCOUNTER = YES;
CLANG_ANALYZER_SECURITY_INSECUREAPI_RAND = YES;
CLANG_ANALYZER_SECURITY_INSECUREAPI_STRCPY = YES;
CLANG_WARN_ASSIGN_ENUM = YES;
CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS = YES;
CLANG_WARN_OBJC_IMPLICIT_ATOMIC_PROPERTIES = YES;
CLANG_WARN_OBJC_INTERFACE_IVARS = YES;
CLANG_WARN_SEMICOLON_BEFORE_METHOD_BODY = YES;
CLANG_WARN_SUSPICIOUS_IMPLICIT_CONVERSION = YES;
GCC_TREAT_WARNINGS_AS_ERRORS = YES;
GCC_WARN_ABOUT_MISSING_FIELD_INITIALIZERS = YES;
GCC_WARN_ABOUT_MISSING_NEWLINE = YES;
GCC_WARN_FOUR_CHARACTER_CONSTANTS = YES;
GCC_WARN_INITIALIZER_NOT_FULLY_BRACKETED = YES;
GCC_WARN_SHADOW = YES;
GCC_WARN_STRICT_SELECTOR_MATCH = YES;
GCC_WARN_UNKNOWN_PRAGMAS = YES;
GCC_WARN_UNUSED_LABEL = YES;
RUN_CLANG_STATIC_ANALYZER = YES;

I'm going to address reported issues (primarily shadow declarations, some dead stores, and a bit more) and submit these all in a PR.

<img width="91" alt="current status" src="https://user-images.githubusercontent.com/7651093/48669570-49efbe80-eabc-11e8-8a4f-98b263615147.png">

bgannin:
Will review and resubmit if applicable for CareKit 2.0.

PR #265, opened 2018-11-21
Project hardening

bgannin:
Turns on additional compiler warnings and analyzer checks per #264. Shallow analyzer is enabled by default now.

Addresses all reported issues so the project builds without warnings and passes both shallow and deep analysis phases. These are enumerated in the individual commits.


srinathtm-apple:

PR #274, opened 2019-06-15
Multi-button log view

pgallastegui:
• Added multi-log task view and its associated controller and bindables
• Refactored code to allow for easier expansion of multi log scenarios (should be easier now to create sliders, custom input, etc).

### DIFF
--- a/CareKit/CareKit.xcodeproj/project.pbxproj
+++ b/CareKit/CareKit.xcodeproj/project.pbxproj
@@ -49,6 +49,9 @@
 		E7D63B1F229D7E6D003A33B4 /* MockSynchronizedViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7D63B1E229D7E6D003A33B4 /* MockSynchronizedViewController.swift */; };
 		E7DA1D5022A30ADC002BB1EF /* OCKListView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7DA1D4F22A30ADC002BB1EF /* OCKListView.swift */; };
 		E7DA1D5322A30AFD002BB1EF /* OCKHeaderBodyView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7DA1D5222A30AFD002BB1EF /* OCKHeaderBodyView.swift */; };
+		F9350B4C22B3F47800CC0399 /* OCKMultiLogTaskViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9350B4A22B3F47800CC0399 /* OCKMultiLogTaskViewController.swift */; };
+		F9350B5522B3FD3D00CC0399 /* OCKBindableMultiLogTaskView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9350B5422B3FD3D00CC0399 /* OCKBindableMultiLogTaskView.swift */; };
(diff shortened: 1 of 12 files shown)


gavirawson-apple:
Thanks for this PR! We love the ideas you have here, and have decided to even take it one step further. We have created a branch `logGrid` where we are building support for a collection view of log buttons. This will be great for anyone that needs more than a single row of log buttons.

Feel free to watch this branch as we work, and contribute any features you would like to see!

pgallastegui:
I'll be checking that branch regularly. So exciting, thank you!

We've tried to expand on this idea in a different branch, creating a new ViewController with a DataSource protocol counterpart that would allow for the users of CareKit to create a custom view that would be in charge to provide a collection of OCKOutcomeValue instances to be logged. This would allow for more granular logging of events like this:

Screen Shot 2019-06-20 at 12 59 01 PM

Following that idea, the completion and data series would have to be based on total quantity logged rather than amount of log entries (leading to a modification of OCKEventAggregator and OCKStoreProtocol.computeCompletion()). Are those areas that you all are planning to explore as well?

erik-apple:
Wow, that's really neat! We're glad to see that you're experimenting with it a lot!
Keep us up to date on your progress, we love to see what you're up to.

You can actually specify your own event aggregators already! There is a second initializer for data series that takes a closure, and within that closure you can implement any logic that you want. This could be used, for example, the sum the number of ounces of breast milk or formula and display how much was consumed on each day.

https://github.com/carekit-apple/CareKit/blob/a82e11bcd369d3408c3cdc37d68a0aadd28a3ba6/CareKit/CareKit/Synchronized%20View%20Controllers/Chart/OCKCartesianChartViewController.swift#L121-L138

pgallastegui:
Excellent! Thank you for pointing me in that direction.

Regarding the daily adherence, at the moment it seems it only considers the count of targetValues passed in the OCKScheduleElement constructor. Is there a way (or are there plans for a way) to make that respond to a custom aggregator?

erik-apple:
If you create an issue requesting that functionality, we'll take it under consideration. It seems like something a fair number of people would want!

scdi:
This is an excellent idea. Good luck with integrating it. Good work.

erik-apple:
@pgallastegui I think this was addressed in another issue somewhere, but for the sake of posterity, I'll document it here as well. You can now pass an aggregator into the constructor of `OCKDailyPageViewController` that will be used to compute adherence values for the completion rings. 

```swift
/// Create an instance of the view controller. Will hook up the calendar to the tasks collection,
/// and query and display the tasks.
///
/// - Parameter storeManager: The store from which to query the tasks.
/// - Parameter adherenceAggregator: An adherence aggregator that will be used to compute the state of the completion rings.
public init(storeManager: OCKSynchronizedStoreManager<Store>, adherenceAggregator: OCKAdherenceAggregator<Store.Event> = .countOutcomes)
```

gavirawson-apple:
@pgallastegui We have updated `master` with changes that address the log button features you requested.

See `OCKButtonLogTaskView` which now has a collection view of log buttons. You can set yourself as the data source to add additional buttons, and even customize the button layout. Further, when one of the buttons is tapped, you can call:

```swift
delegate?.eventView(self, didCreateOutcomeValueAt: indexPath.row, sender: cell)
```

 to notify the view controller to create a new outcome value in the store. Essentially, that allows you to create a custom view without having to worry about creating or modifying a view controller!

I believe this solves the issues you raised in this PR, but let us know if this PR still applies!

PR #295, opened 2019-07-01
Feature & Sample App Improvements; Bug Fixes

oalejel-apple:
* New OCKTaskQuery property to exclude tasks that do not have active events in the queried time interval.
* Sample app demonstrates better example of displaying tasks. 
* TipView in sample app only appears on current day. 
* Bug fixes related to week page scrolling and other UI issues.

### DIFF
--- a/CareKit/CareKit/Lists/Controller/OCKDailyPageViewController.swift
+++ b/CareKit/CareKit/Lists/Controller/OCKDailyPageViewController.swift
@@ -128,6 +128,7 @@ UIPageViewControllerDelegate {
         let now = Date()
         calendarPageViewController.calendarDelegate = self
         calendarPageViewController.selectDate(now, animated: false)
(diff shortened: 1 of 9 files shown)


CareKit/CareKit/Lists/Controller/OCKDailyPageViewController.swift
```diff
@@ -128,6 +128,7 @@ UIPageViewControllerDelegate {
         let now = Date()
         calendarPageViewController.calendarDelegate = self
         calendarPageViewController.selectDate(now, animated: false)
+        view.backgroundColor = OCKStyle.color.gray1 // avoid black background flashing
```
srinathtm-apple: I think we should remove this comment...

srinathtm-apple: nvm, not a big deal..

PR #311, opened 2019-09-05
add detail disclosure to simple log view

jarrodldavis:
fix the header view's detail disclosure for the simple log task view

### DIFF
--- a/CareKit/CareKit/Synchronized View Controllers/Task/OCKSimpleLogTaskViewController.swift
+++ b/CareKit/CareKit/Synchronized View Controllers/Task/OCKSimpleLogTaskViewController.swift
@@ -42,6 +42,10 @@ open class OCKSimpleLogTaskViewController<Store: OCKStoreProtocol>: OCKEventView
         return view
     }
 
+    override var detailPresentingView: UIView? {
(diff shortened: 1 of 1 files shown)


gavirawson-apple:
Thanks for the PR! We decided not to show a detail view controller when the simple task card is tapped. Rather when tapped, the event in the card will be marked complete/incomplete. That functionality is not there now, but will be added very soon when we push a bunch of new changes. 

From a design perspective, we felt that the card was too small to have both of these tap behaviors. We also felt that some people may find the circle button in the card too small of a tap area. So, we decided to go the route that makes the card more accessible. 

How do you feel about that design decision? Does it make sense to you and your users from a UI/UX perspective?

jarrodldavis:
I'm not sure I understand. This is a change for the `simpleLog` view, which looks like this:

Screen Shot 2019-09-09 at 5 31 52 PM

What you're describing (with the small size and circle button) sounds like the `simple` view:

Screen Shot 2019-09-09 at 5 33 47 PM

gavirawson-apple:
Apologies, you're absolutely right! Thank you for the fix!

PR #330, opened 2019-12-04
Change XCode to Xcode

stevemoser:
Thank you for making this awesome framework! I should have caught this and added it to the other PR. I think this will be the last typo fix for a bit.


erik-apple:
They’re always welcome! Thanks for helping us find these typos.

Issue #355, opened 2020-01-31
[Proposal] Category Card, OCKTaskCategory, organization of OCKTasks


kutakmir:
I'd like to propose a card component that will enable us to organize tasks by category. Something similar to the Apple Health app's Browse panel. For that purpose, there would be a new entity -> OCKTaskCategory. Currently, there is a groupIdentifier in the OCKTask that fits this purpose, but that doesn't allow for efficient querying of all the available categories.

The OCKTaskCategory will have a title: String and an icon: UIImage properties. The design will be very similar to the Apple Health app.

In addition to this card + model layer stack (CareKit Store and Core Data), there could be a ViewController portion with a search bar and index bar.

....

Then a similar pattern can be used for displaying Care Plans as they are another, currently supported, way of organizing OCKTasks.

This is something that I can implement and submit a PR.

What do you think? Is this a desirable addition to CareKit?

pthealthtech:
This functionality would be very beneficial.

gavirawson-apple:
This sounds like a great idea! If you would like to work on this feature, we would be happy to merge it into the framework. Otherwise we can build it out on our end. 

For the Core Data step, it will be helpful to look at the way notes are set up in the schema. There is one note entity with potential relationships to the owner of the note. You can setup a new `OCKCategory` entity in the same manner, which will allow you to perform a query such as "Give me all categories that belong to a task." Going this route will save you from creating a separate category for each entity (`OCKTaskCategory`, `OCKCarePlanCategory`, etc).

The view controller portion sounds great as well. It might be a good route to query and display each category, and allow the developer to define what to do when the category is tapped (perhaps through a delegate method). That way, the view controller will be flexible for a variety of use cases.

If you're feeling super ambitious, you could take a shot at keeping the displayed categories synchronized with the ones in the database (I.E if a category is removed from the database, it should no longer be displayed in the view). This will be a bit challenging - we're happy to either work with you on this part or append that functionality for you to your PR.

PR #386, opened 2020-03-16
Implemented TaskCategory entity #355

kutakmir:
Addresses issue #355.

A new TaskCategory entity was introduced. This feature helps to organize OCKTasks in categories.
This way the user can categorize the tasks by a medical condition, by the type of action, or in any other way they want.

### New entities
```
TaskCategory
...
- title
- asset (image name)
- notes array (contains the list of textual notes created by the practitioners or the patient)
```
### New relationships
```
TaskCategory -> CarePlan
CarePlan ->> TaskCategory

TaskCategory ->> Task
Task -> TaskCategory
```

```
-> has one
->> ... has many
```

## Considerations

1) Shouldn't we add an "Uncategorized" TaskCategory that will always be there and all the Tasks that are not associated with any TaskCategory will be displayed when this  "Uncategorized" TaskCategory is selected?
...
Currently, the taskCategory property is optional. To implement the "Uncategorized" TaskCategory would mean adding a special logic for this use-case. A more generic solution is to solve this issue upon task creation. If you create a new task, you'll choose a category. In no category is selected, there will be the default "Uncategorized" TaskCategory. But this logic is something to be decided by the end-user (the programmer of a particular app). Some apps might not want to categorize the tasks, so the default category would render no benefit.

2) Nesting of the categories
...
It's a bit tricky because Swift structs cannot be nested directly (to prevent infinite loop upon initialization). The alternative is to use IDs. Is this something that is desirable? What is the best way to achieve nesting?

3) Gravirawson's suggestion https://github.com/carekit-apple/CareKit/issues/355#issuecomment-584300892: create an OCKCategory instead of OCKTaskCategory for CarePlans and other use-cases.
...
What would be the use-case for the care plan category? Could be the same category reused in multiple different entities?

4) Should the TaskCategory to Task relationship be many-to-many instead?
...
This would provide more flexibility. Right now, I only have one task category per task for simplicity's sake. The many-to-many would enable to filter tasks by different sets of categories. This is especially useful for the implementation of the "All tasks" category (that we can include in the list of categories).
```
TaskCategory ->> Task
Task ->> TaskCategory
```
5) What is the best way to integrate this new functionality in terms of UI?
...
Really an open question for debate. I created a very basic category selection view controller that in turn presents the daily list of tasks filtered by a task category. Should the filtering happen on the main screen? Or would a completely different UI be more appropriate? The original idea was to create the same drill-down experience as in the Apple Health app.

6) Should we keep the notes array in the OCKTaskCategory?


erik-apple:
Thanks for the PR! We'll begin going over it and leaving feedback soon!

CareKit/CareKit/Lists/Controller/OCKTaskCategoriesListViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override open func viewDidLoad() {
+        super.viewDidLoad()
+        navigationController?.navigationBar.prefersLargeTitles = true
+        title = loc("TASKCATEGORIES")
```
erik-apple:
```suggestion
        title = loc("TASK_CATEGORIES")
```

CareKit/CareKit/Synchronized View Controllers/TaskCategory/ViewControllers/OCKDetailedTaskCategoryViewController.swift
```diff
@@ -0,0 +1,58 @@
+//
```
erik-apple: nitpick: Remove the first `//` at the top of new files.

CareKit/CareKit/Synchronized View Controllers/Utilities/OCKTaskCategoryUtility.swift
```diff
@@ -0,0 +1,48 @@
...
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import MapKit
```
erik-apple: Is there a reason MapKit is imported here?

CareKit/CareKitTests/Contact/TestSimpleContactViewSynchronizer.swift
```diff
@@ -85,6 +85,6 @@ class TestSimpleContactViewSynchronizer: XCTestCase {
         viewSynchronizer.updateView(view, context: .init(viewModel: nil, oldViewModel: nil, animated: false))
         XCTAssertNil(view.headerView.titleLabel.text)
         XCTAssertNil(view.headerView.detailLabel.text)
-        XCTAssertEqual(view.headerView.iconImageView?.image, UIImage(systemName: "person.crop.circle"))
+        XCTAssertEqual(view.headerView.iconImageView?.image, UIImage(systemName: "questionmark.circle.fill"))
```
erik-apple: I think we may want to keep the default image as "person.crop.circle"

CareKitStore/CareKitStore/CoreData/OCKCDTaskCategory.swift
```diff
@@ -0,0 +1,62 @@
...
+
+private extension Array where Element == OCKLabeledValue {
+    func asDictionary() -> [String: String] {
+        var dictionary = [String: String]()
+        for labeldValue in self {
+            dictionary[labeldValue.label] = labeldValue.value
+        }
+        return dictionary
+    }
+}
```
erik-apple: These extensions don't appear to be being used. Is it possible to remove them?

CareKitStore/CareKitStore/CoreData/OCKManagedObjectModel.swift
```diff
@@ -678,6 +756,28 @@ private func makeContactEntity() -> NSEntityDescription {
...
+    let title = NSAttributeDescription()
+    title.name = "title"
+    title.attributeType = .stringAttributeType
+    title.isOptional = true
+
+    let other = NSAttributeDescription()
+    other.name = "otherTaskCategoryInfoDictionary"
+    other.attributeType = .transformableAttributeType
+    other.isOptional = true
+    other.valueTransformerName = secureUnarchiver
```
erik-apple: What is this attribute intended to be used for? I don't think I saw an `@NSManaged` variable for it in `OCKCDTaskCategory`.

CareKitUI/CareKitUI/Components/TaskCategory/Button/OCKTaskCategoryButton.swift
```diff
@@ -0,0 +1,153 @@
...
+        var image: UIImage? {
+            let image: UIImage?
+            switch self {
+            case .call: image = UIImage(systemName: "phone")
+            case .message: image = UIImage(systemName: "text.bubble")
+            case .email: image = UIImage(systemName: "envelope")
+            }
+            return image
+        }
+    }
```
erik-apple: Why does the task category button have this code in it?

CareKitUI/CareKitUI/Common/Views/OCKTaskCategoryHeaderView.swift
```diff
@@ -0,0 +1,237 @@
...
+///    | +----+                                 |
+///    | |icon|  Title             [detail      |
+///    | |img |  Detail             disclosure] |
+///    | +----+                                 |
+///    |                                        |
+///    |  ------------------------------------  |
+///    |                                        |
+///    +----------------------------------------+
+///
+open class OCKTaskCategoryHeaderView: OCKView {
```
erik-apple: Can you explain to us how the view is different from the normal `OCKHeaderView`?

CareKitUI/CareKitUI/Supporting Files/Localization/en.lproj/Localizable.strings
```diff
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 "CALL" = "Call";
 "COMPLETED" = "Completed";
 "CONTACTS" = "Contacts";
+"TASKCATEGORIES" = "Task Categories";
```
erik-apple:
```suggestion
"TASK_CATEGORIES" = "Task Categories";
```

CareKitUI/CareKitUI/Common/Views/OCKTaskCategoryHeaderView.swift
kutakmir: At the moment, it's not any different. I thought of a different appearance but didn't really pursue it. A better practice would be to parametrize the OCKHeaderView. I'll replace the OCKTaskCategoryHeaderView reference with OCKHeaderView.

CareKitUI/CareKitUI/Components/TaskCategory/Button/OCKTaskCategoryButton.swift
kutakmir: Let me remove the whole class, I don't think we'll need such a button.

CareKitStore/CareKitStore/CoreData/OCKManagedObjectModel.swift
kutakmir: Forgot to remove this one. Removing now...

CareKitStore/CareKitStore/CoreData/OCKCDTaskCategory.swift
kutakmir: Yes

CareKit/CareKitTests/Contact/TestSimpleContactViewSynchronizer.swift
kutakmir: Yes, and we are using that. I introduced the questionmark icon there so that we can use it as a universal placeholder.

If you look at OCKHeaderView+Updatable, you'll see the updated logic. We are using OCKHeaderView for various purposes - Events, Tasks, Contacts, and TaskCategories. If the clear out and not configured state was using a person placeholder, it would look strange in the TaskCategory list.

The Contact list will not be impacted by this change. The difference is in the configuration. If you configure the OCKHeaderView with a OCKAnyContact that doesn't have any asset - the image will be rendered as a person placeholder. In case of the OCKAnyTaskCategory, a question mark image will be used.

Let me know if this is desirable logic or what other solution would you suggest.

CareKit/CareKit/Synchronized View Controllers/Utilities/OCKTaskCategoryUtility.swift
kutakmir: No, I will remove it.

CareKit/CareKit/Synchronized View Controllers/TaskCategory/ViewControllers/OCKDetailedTaskCategoryViewController.swift
kutakmir: Done

CareKit/CareKit/Lists/Controller/OCKTaskCategoriesListViewController.swift
kutakmir: done

CareKitUI/CareKitUI/Supporting Files/Localization/en.lproj/Localizable.strings
kutakmir: Fixed.

kutakmir:
@erik-apple for some reason the CI is giving me a failure in some of the unit tests, although they pass when I run them locally. Are you able to reproduce?

    ✗ testTaskQueryOnPastDateReturnsPastVersionOfATask, XCTAssertTrue failed - Expected to get 1 task, but got 0
    ✗ testTaskQueryOnPastDateReturnsPastVersionOfATask, XCTAssertTrue failed

erik-apple:
Yes, that’s a known issue. We’ve noticed that it happens at a certain time of day, and only on Travis’s CI servers. We are unable to reproduce it locally as well.

If you wait a couple hours and rerun the tests they will pass. Don’t worry about it for now though, we’ll ensure they pass before we merge.

kutakmir:
@erik-apple Hi Erik, it's been awhile, would you mind giving me some guidance so that I can make all required alterations to get this PR merged? Thanks in advance!

gavirawson-apple:
I'm closing this out for now because the main branch has changed quite a bit since this was opened. If you still think this feature should be added to the codebase, we can use this thread to discuss the implementation details.

PR #387, opened 2020-03-16
Add completion to log button

tommysarni:
I need a new View Controller to pop up when pressing the Log Button, but still like the idea of the logging when the button was pressed. In order to keep protection, I simply created a function property for the OCKButtonLogTaskView.

### DIFF
--- a/CareKitUI/CareKitUI/Components/Task/OCKButtonLogTaskView.swift
+++ b/CareKitUI/CareKitUI/Components/Task/OCKButtonLogTaskView.swift
@@ -89,6 +89,7 @@ open class OCKButtonLogTaskView: OCKLogTaskView, UICollectionViewDelegate, UICol
         return label
     }()
     
+    /// Completion after log button is pressed
     public var logPressedCompleted: (() -> ())?
 
     // MARK: Methods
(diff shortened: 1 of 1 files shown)


tommysarni:
@erik-apple Can you review this PR and if this is not possible can you please guide me in how to create my own TaskViewController. Thanks so much

erik-apple:
I think subclassing is the route you should probably go! 

Try out something like this.

```swift
class TommyTaskViewController: OCKButtonLogTaskViewController {
    
    override func taskView(_ taskView: UIView & OCKTaskDisplayable,
                           didCompleteEvent isComplete: Bool,
                           at indexPath: IndexPath, sender: Any?) {
        
        // Do what the superclass does (log new outcome, delete outcome)
        super.taskView(taskView, didCompleteEvent: isComplete, at: indexPath, sender: sender)
        
        // Plus insert your extra code here (only when completing a new outcome)
        if isComplete {
            let someViewController = UIViewController()
            present(someViewController, animated: true, completion: nil)
        }
    }
}

tommysarni:
Awesome, I'll try it out and let you know how it goes !

erik-apple:
@tommysarni How did that work out? Is it okay to close this PR?

tommysarni:
@erik-apple Oh so sorry totally forgot to update, yeah subclassing worked perfectly. Thanks so much !

erik-apple:
Great! I'll close this out then!

PR #389, opened 2020-03-18
Fix umbrella framework

gavirawson-apple:
CareKit is no longer an umbrella framework that embeds both CareKitUI and CareKitStore.

Note: If you have installed CareKit by embedding the framework, make sure that your app embeds CareKit, CareKitUI, and CareKitStore.


erik-apple:
When this goes live on GitHub, let's put instructions about how to configure projects properly in the PR so we can point people to it when the inevitably start having issues.

PR #403, opened 2020-04-06
Fix memory leak

gavirawson-apple:
Fixes a memory leak that stopped some views from getting deallocated.


erik-apple (approved):

kristinalynettetyree-collab:
@erik-apple [__]()

PR #435, opened 2020-05-20
Query by uuid for Outcome and expose deletedDate

### LINKED ISSUE #432: `OCKOutcome.deletedDate` is marked internal
All of the other entities have this as public

### LINKED ISSUE #433: OCKOutcome.uuid is marked internal and can't query by OCKOutcome.uuid
- `OCKOutcome.uuid` is marked internal. This might be because of the top-level and version entity discussion we had earlier. The problem here is during syncing, if the uuid isn't pushed to the cloud, I don't see how it can be pulled in pullRevision. *Workaround* I had to resort encoding/decoding of `OCKOutcome` which I later found is required for an `OCKEntity` that is part of a `OCKRevisionRecord`. So this might be the way CareKit wants us to go. I can see how json makes for versatility of exporting/importing. In ParseCareKit, I moved to encoding/decoding the bare minimum, i.e. uuid, createdDate, updatedDate for OCKEntity, and set the mutable types after the entity is created.
- Probably related to above, but there's no way to query`OCKOutcome` by `uuid`. My workaround is to add `tag` by the `entityId` I added to `userInfo` and query by that

cbaker6:
- Fix #432 
- Fix #433
- Fix query by uuids for OCKContact
- Fix query by patient uuids for OCKCarePlan
- Slight change to mapping ids and uuids in OCKCoreDataStoreProtocol.swift so that CareKit can still compile on < Xcode 11.4.0
- Downgrade travis to xcode11.3 and iOS13.3 since xcode11.4 is having issues, see [this](https://github.com/carekit-apple/CareKit/pull/435#issuecomment-632079613) for details.

### DIFF
--- a/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
+++ b/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
@@ -223,7 +223,6 @@ class TestStoreOutcomes: XCTestCase {
         task = try store.addTaskAndWait(task)
 
         var outcome = OCKOutcome(taskUUID: try task.getUUID(), taskOccurrenceIndex: 0, values: [])
-        outcome.tags = ["123"]
         outcome = try store.addOutcomeAndWait(outcome)
 
         var query = OCKOutcomeQuery(for: Date())
--- a/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
+++ b/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
@@ -217,6 +217,24 @@ class TestStoreOutcomes: XCTestCase {
         let fetched = try store.fetchOutcomesAndWait(query: query).first
         XCTAssert(fetched == outcome)
     }
+ 
+    func testQueryOutcomeByUUID() throws {
+        var task = OCKTask(id: "A", title: nil, carePlanUUID: nil, schedule: .mealTimesEachDay(start: Date(), end: nil))
+        task = try store.addTaskAndWait(task)
+
+        var outcome = OCKOutcome(taskUUID: try task.getUUID(), taskOccurrenceIndex: 0, values: [])
+        outcome.tags = ["123"]
+        outcome = try store.addOutcomeAndWait(outcome)
+
+        var query = OCKOutcomeQuery(for: Date())
+        query.tags = ["123"]
+
+        let fetched = try store.fetchOutcomesAndWait(query: query).first
+        var query2 = OCKOutcomeQuery(for: Date())
+        query2.uuids = [fetched!.uuid!]
+        let fetched2 = try store.fetchOutcomesAndWait(query: query2).first
+        XCTAssert(fetched2 == outcome)
+    }
 
     // MARK: Updating
 
--- a/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
+++ b/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
@@ -227,14 +227,10 @@ class TestStoreOutcomes: XCTestCase {
         outcome = try store.addOutcomeAndWait(outcome)
 
         var query = OCKOutcomeQuery(for: Date())
-        query.tags = ["123"]
-
-        let fetched = try store.fetchOutcomesAndWait(query: query).first
-        var query2 = OCKOutcomeQuery(for: Date())
-        query2.uuids = [fetched!.uuid!]
+        query.uuids = [ outcome.uuid!]
         
-        let fetched2 = try store.fetchOutcomesAndWait(query: query2).first
-        XCTAssert(fetched2 == outcome)
+        let fetched = try store.fetchOutcomesAndWait(query: query).first
+        XCTAssert(fetched == outcome)
     }
 
     // MARK: Updating
--- a/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
+++ b/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
@@ -232,6 +232,7 @@ class TestStoreOutcomes: XCTestCase {
         let fetched = try store.fetchOutcomesAndWait(query: query).first
         var query2 = OCKOutcomeQuery(for: Date())
         query2.uuids = [fetched!.uuid!]
+        
         let fetched2 = try store.fetchOutcomesAndWait(query: query2).first
         XCTAssert(fetched2 == outcome)
     }
--- a/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
+++ b/CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
@@ -227,7 +227,7 @@ class TestStoreOutcomes: XCTestCase {
         outcome = try store.addOutcomeAndWait(outcome)
 
         var query = OCKOutcomeQuery(for: Date())
-        query.uuids = [ outcome.uuid!]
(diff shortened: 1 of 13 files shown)


cbaker6:
I think travis had an issue with building as it didn't make it to the test

erik-apple:
Thanks for the PR! I noticed travis builds have all been failing recently. I'm looking into the cause!
We'll get this merged in as soon as travis is working again!

cbaker6:
@erik-apple Looks like travis may have removed 13.4, travis works now with 13.4.1

TestStoreBuildRevisions is what's failing, but I think it was a random timeout because travis was taking forever. It passes for me locally. Are you able to rerun travis?

erik-apple:
Yep, I'll try kicking it off again.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
```diff
@@ -217,6 +217,25 @@ class TestStoreOutcomes: XCTestCase {
...
+        var outcome = OCKOutcome(taskUUID: try task.getUUID(), taskOccurrenceIndex: 0, values: [])
+        outcome.tags = ["123"]
+        outcome = try store.addOutcomeAndWait(outcome)
+
+        var query = OCKOutcomeQuery(for: Date())
+        query.tags = ["123"]
+
+        let fetched = try store.fetchOutcomesAndWait(query: query).first
+        var query2 = OCKOutcomeQuery(for: Date())
+        query2.uuids = [fetched!.uuid!]
```
erik-apple, line 234: Is the reason you're tagging the outcome and querying by tag so that you can get the `uuid` of the persisted outcome?

If so, then you can skip querying. The UUID is already populated on line 227. You can just grab `outcome.uuid`.

cbaker6: Right, I'll make the change

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
```diff
@@ -227,14 +227,10 @@ class TestStoreOutcomes: XCTestCase {
         outcome = try store.addOutcomeAndWait(outcome)
 
         var query = OCKOutcomeQuery(for: Date())
-        query.tags = ["123"]
-
-        let fetched = try store.fetchOutcomesAndWait(query: query).first
-        var query2 = OCKOutcomeQuery(for: Date())
-        query2.uuids = [fetched!.uuid!]
+        query.uuids = [ outcome.uuid!]
```
erik-apple, line 230: Looking good! The last change I'll ask you to make is to change `outcome.uuid!` to `try outcome.getUUID()`. If you look at some of the other tests, you'll see that we try to prefer using throwing getters over force unwraps in the context of unit tests.

Our reasoning is that should something happen to go wrong, we want the test to fail without crashing so that we can pinpoint the regression. If a force unwrap crashes the test runner, we won't be able to tell where the test failed.

Look around for a test extension on `OCKTask` that gives it the `func getUUID() throws -> UUID` method we use in tests and add another similar method for `OCKOutcome`, then employ that here.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
```diff
@@ -217,6 +217,21 @@ class TestStoreOutcomes: XCTestCase {
         let fetched = try store.fetchOutcomesAndWait(query: query).first
         XCTAssert(fetched == outcome)
     }
+ 
+    func testQueryOutcomeByUUID() throws {
+        var task = OCKTask(id: "A", title: nil, carePlanUUID: nil, schedule: .mealTimesEachDay(start: Date(), end: nil))
+        task = try store.addTaskAndWait(task)
+
+        var outcome = OCKOutcome(taskUUID: try task.getUUID(), taskOccurrenceIndex: 0, values: [])
+        outcome.tags = ["123"]
```
erik-apple, line 226: Oh, actually, you can remove these tags now as well!

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Outcomes.swift
cbaker6: Added, but question. 

I added the getUUID extension on `OCKObjectCompatible` because the other versioned objects seemed to be getting it from `OCKVersionedObjectCompatible`. If this is okay, it would seem the extension `OCKVersionedObjectCompatible` can be removed because all versioned objects are `OCKObjectCompatible`.

I believe this enables getUUID on OCKNote and OCKValue and the like as well, which I'm not sure if that should be the case

erik-apple: Yeah, you're 100% right. I think the right route is to just change the existing extension to be on `OCKObjectCompatible`. Previous only versioned objects had uuids, but now that outcomes do too, it makes sense to expand that scope of that extension!

cbaker6: I'm sure it will pass, only thing now is a slight naming convention.

So in files in `buildPredicate` in versioned object files like OCKStore+CarePlans.swift, they use `versionPredicate` instead of `objectPredicate`. It still works of course:

```swift
if !query.uuids.isEmpty {
            let versionPredicate = NSPredicate(format: "%K IN %@", #keyPath(OCKCDVersionedObject.uuid), query.uuids)
            predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [predicate, versionPredicate])
        }
```

Should I change them or leave them?

erik-apple: Might as well change them while we're at it!

erik-apple: Thanks for boy-scouting!

cbaker6: I'm think I may have found a bug in OCKStore+CarePlans.swift while changing these:

```
if !query.patientUUIDs.isEmpty {
            let objectPredicate = NSPredicate(format: "%K IN %@", #keyPath(OCKCDCarePlan.patient), query.patientUUIDs)
            predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [predicate, objectPredicate])
        }
```

I think that keypunch should be `OCKCDCarePlan.patient.uuid` instead of `OCKCDCarePlan.patient`

cbaker6: Another one in contacts. These are probably here because of the conversion to UUIDs:

```
if !query.uuids.isEmpty {
            let versionPredicate = NSPredicate(format: "%K IN %@", #keyPath(OCKCDVersionedObject.id), query.uuids)
            predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [predicate, versionPredicate])
        }
```

That should now be  `OCKCDObject.uuid `

cbaker6:
Both of the tests that failed in travis pass locally for me. Are you able to rerun?

cbaker6:
I was looking around on travis forums and it seems the Xcode 11.4 image has been having issues. They recently updated to 11.4.1 (11.4 defaults to this now), but it's still buggy. Interestingly enough, when I told travis to use xcode11.4.1, a different test failed, then what was previously failing.

cbaker6:
I was looking around on travis forums and it seems the Xcode 11.4 image has been having issues. They recently updated to 11.4.1 (11.4 defaults to this now), but it's still buggy.

I reverted back to 11.3.1 and applied a small fix in OCKCoreDataStoreProtocol.swift that was preventing compiling on < Xcode 11.4.0.

Issue #418, opened 2020-04-21
Question about OCKOutcome, OCKOutcomeValue, OCKSchedule, and OCKScheduleEvent


cbaker6:
It seems that OCKOutcome, OCKOutcomeValue, OCKSchedule, and OCKScheduleEvent are setup slightly different from OCKPatient, OCKCarePlan, OCKTask, etc. For example: 

1. OCKOutcome cannot be initialized with with an `id` and OCKStore creates one on it's own that looks something like, `x-coredata://BC6D5810-5D8F-4E27-A74B-C33F32E19387/OCKCDTask/p22_7` which looks more like the `localDatabaseID`
2. OCKOutcome, OCKOutcomeValue, OCKSchedule, and OCKScheduleEvent doesn't have an `id` at all
3. OCKSchedule doesn't have a `remoteId` (I haven't needed this so far because I've only needed to sync OCKScheduleElement to the cloud)

To get around the issues above, I use `userInfo` to add UUID's for each. Are id's missing, or in the case of OCKOutcome having it's `id` used in a different way setup like this for a particular reason?

The reason I ask this because when syncing between local devices and cloud, if a local OCKStore entity doesn't have an `id` and it's `remoteId` isn't set, then I don't see a way to link the local and cloud entities

erik-apple:
All very good questions and observations!

We like to describe the entities in CareKit’s store using the terms “top level”, “versioned”, and “un-versioned”.

Top level entities are the ones that you can save into the store directly. The top level entities are OCKPatient, OCKCarePlan, OCKContact, OCKTask, and OCKOutcome. By contrast, OCKSchedule, OCKOutcomeValue, and OCKPostalAddress (to name a few) can only be inserted into the store by attaching them to a top level entity and saving that top level entity into the store.

Versioned entities are just that — entities that are stored as a linked list of versions.
The versioned entities are OCKPatient, OCKCarePlan, OCKContact, and OCKTask.

Un-versioned entities do not have a notion of versions. The most notable of these is OCKOutcome because it’s also a top level entity. The others (OCKSchedule, OCKOutcomeValue, OCKPostalAddress, etc.) are pretty boring. 

The versioned entities can all be initialized with an `id`. I think you understand the pattern pretty well already. The `id` is used to associate different versions of the same object.

[...]

Issue #416, opened 2020-04-20
Syncing data to the cloud with remoteID


cbaker6:
Hi,

Is there a recommended way to sync data to the cloud based on `remoteID`. I ask this because the documentation states, `remoteID: A string value that represents a unique identifier, which a remote database uses optionally.`.

What I currently do is after an entity is synced to the cloud, lets say OCKPatient, is take the `objectId` (UUID generated for the cloud object) as the `remoteID` of OCKPatient. Which seems to work fine.

I use `Combine` and `OCKSynchronizedStoreManager` to subscribe to changes like add and updated and attempt to sync to the cloud as recommended by @erik-apple on a previous issue. Of course, there can be issues during a push to the cloud, or no connectivity at all. So whenever there's connectivity available, I would like to search the OCKStore for OCPatient where remoteID == nil. Basically, any entity that hasn't received a remoteID is considered not in the cloud.

1. Is this the intended use of remoteID or at least a good way to use it?
2. If this is a good way, is there a way to fetch `remoteID==nil` (and any other entity that can be nil)? This would be straightforwad if I was using NSPredicate directly to fetch from the OCKStore, but I can't figure out how to do this with the current fetch methods. My current workaround is to fetch all entities, and then filter by remoteID below:

```
fileprivate func synchronizePatients(){
        let query = OCKPatientQuery(for: Date())
        
        storeManager.store.fetchAnyPatients(query: query, callbackQueue: .main){
            result in
            
            switch result{
                
            case .success(let foundPatients):
                
                guard let patients = foundPatients as? [OCKPatient] else{return}
                
                let patientsToSync = patients.filter{$0.remoteID == nil}
                self.addCloudPatients(patientsToSync)
                
            case .failure(let error):
                print("Error in synchronizePatients(). \(error)")
            }
        }
    }
``` 

I haven't ran into an issue with my current way, it just seems like I should be able to query the OCKStore instead of using the filter after getting all entities.

I suppose one way I can get around this is whenever I create a new OCKPatient, set the remoteID to a known string value, like "undefined" or "notInCloud" and then set `query.remoteID = ["notInCloud"]`, then fetch

erik-apple:
Yes, your approach is correct.

Each version of a patient is immutable, so it is sufficient to simply check if that version has been uploaded already or not. Since the patient cannot change without adding a new version, we don't need to consider the update case. Deletes, likewise, are modeled by adding a new version with the `deletedDate` set to a non-nil value.

[...]

PR #436, opened 2020-05-21
Allow queries that match against nil

erik-apple:
This PR makes it possible to query objects where properties are `nil`.

This is useful for returning, for example, all tasks that don't have a `groupIdentifier` or all outcomes for which the `remoteID` is not set.

This feature was proposed by @cbaker6 in #418.


cbaker6:
Cool! I think I mentioned this in #416 also

PR #443, opened 2020-06-04
Ensure that the graph area grows before the legend when stretched

erik-apple:
This patch ensures that the graph area in `OCKCartesianChartView` grows first when the view is constrained to a size greater than its intrinsic content height.


scottmtp:
FYI, it appears that the OCKSample app reports a number of LayoutConstraints problems after this was merged.

PR #448, opened 2020-06-07
Update logicalClock of tombstoned and previous version entities

### LINKED ISSUE #446: Handling the remote synchronization of a deleted Outcome
Hi @erik-apple, I've been looking at the tombstoning of an outcome when a "delete" occurs and it doesn't seem like the tombstoned outcome in `updateOutcomesLeavingTombstone`:

```swift
currentOutcomes.forEach {
            $0.deletedDate = Date()
            $0.values = Set()
}
```

gets added to the revision record. I do see the new outcome in the revision record with a deletedDate set, but this leaves the original outcome with no deletedDate on the remote still referencing it's outcomeValue. So on the remote, the original outcome is still active, though in the OCKStore it's correctly tombstoned.

I'm guessing the `logicalClock` of the tombstoned outcome isn't being incremented when it's tombstoned, so `changedQuery` isn't picking it up, but I couldn't find where `logicalClock` should be updated

cbaker6:
This ensures all synchronized entities that are updated/deleted are added to the upcoming RevisionRecord so their changes are pushed to the remote.

Fix #446


CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
```diff
@@ -267,6 +267,40 @@ class TestStoreSync: XCTestCase {
         XCTAssert(localOutcomes == remoteOutcomes)
         XCTAssert(localOutcomes.count == 1)
     }
+    
+    func testTombstoningOutcomePushedToRemote() throws {
+        let testStore = OCKStore(name: "test", type: .inMemory, remote: dummy)
+        dummy.automaticallySynchronizes = false
```
cbaker6, line 273: So this test should really be for when automaticallySynchronizes = true, but I couldn't figure out how to test this properly.

The problem it's trying to test is:

1) add an outcome to the local store
2) sync that outcome to remote
3) delete the outcome from the local store causing a tombstone
4) the updated tombstoned and the new outcome replacing the previous are synced to the remote

When using `syncAndWait` it seems to synchronize all entities again, in this test case, 3 entities, where there should only be 2

cbaker6:
I’m not sure how adding a test case caused two of the task test cases to start failing.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: I got around the previous issue by adding a new DummyEndpoint (DummyEndpoint2) which increments and merges it’s knowledgeVector which is needed to test the problem of this fix. Running the testcase on previous versions without the fix will fail

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
```diff
@@ -267,6 +267,47 @@ class TestStoreSync: XCTestCase {
...
+            case .outcome(let outcome):
+                return outcome
+            default:
+                return nil
+            }
+        }
+        XCTAssert(tombstonedOutcomes.count == 2)
+        
+        guard let tombstonedWithSameUUID = tombstonedOutcomes.filter({$0.uuid == outcome.uuid}).first else{
+            throw OCKStoreError.invalidValue(reason: "Filter doesn't contain UUID")
```
cbaker6, line 300: Wasn’t sure what to throw here, if there’s a suggested way, I can change

CareKitStore/CareKitStoreTests/OCKStore/TestStore+BuildRevisions.swift
```diff
@@ -94,7 +94,7 @@ class TestStoreBuildRevisions: XCTestCase {
         try store.deleteTasksAndWait([taskA2])
         let revision = store.computeRevision(since: store.context.clockTime)
 
-        XCTAssert(revision.entities.count == 1)
+        XCTAssert(revision.entities.count == 2)
```
cbaker6, line 97: This was revealed after fixing the logicalClock on a versioned outcome. It seems like it should be 2, as the added entity and the deleted entity should both be pushed to the remote

cbaker6:
It looks like the 2 added tests could have possibly been added to `TestStoreBuildRevisions` instead of `TestStoreSync` since the problem was populating the RevisionRecord to be synced to the remote. Since these tests use `syncAndWait` along with testing if the remote gets the correct revisions, are they okay here? I see the other way would be to use `store.computeRevision()` instead of DummyEndpoint2

CareKitStore/CareKitStore/Protocols/CoreData/OCKCoreDataStoreProtocol.swift
```diff
@@ -180,6 +180,8 @@ extension OCKCoreDataStoreProtocol {
             let newVersion = addNewVersion(value)
             newVersion.previous = current
             newVersion.uuid = UUID()
+            current.next = newVersion
```
cbaker6, line 183: I couldn't find in the code where "next" was being set. This seemed like the appropriate place. `testUpdateTaskVersionPushedToRemote()` also tests to see if next is pushed to the remote

cbaker6:
@erik-apple it looks like this may have been fixed in the updates in a way that doesn't update the logical clock, but still syncs the delete/tombstone?

If so, I'm not sure if the test cases in this PR may still be be helpful

cbaker6:
I deleted my previous post as it looks like I was wrong about the problem being fixed. I have checked this merge locally and it works.

I have an example using the WWDC20 OCKSample with the AppleWatch code that should show a delete synchronizing across watch and phone, that I'll post in few...

cbaker6:
Here's the [latest](https://github.com/netreconlab/CareKitSample-ParseCareKit/tree/experimental-2.1) OCKSample app using the this PRs fixes. It works a little different than the WWDC video as the watch and the iOS are both using ParseCareKit to sync through the cloud remote (parse-hipaa running locally on your computer via docker). The tombstoning of tasks from watch to iPhone works the same as the video (the watch pings (through WCSession) the iOS device to sync with the parse-hipaa remote whenever it successfully pushes to the remote). Changes from the iPhone to watch can't be seen unless you exit the watch app and comeback (for some reason I can't get the iPhone to talk back to the watch via WCSession. Not sure if it's the beta I'm using or if I setup something wrong with WCSession).

The version of the sample app using the current CareKit branch is [here](https://github.com/netreconlab/CareKitSample-ParseCareKit/tree/main). Visually, this looks the same as deletes will disappear as I mentioned earlier, but the tombstoning of the original outcome on the remote doesn't happen which can be seen on the parse-hipaa dashboard. 

I couldn't get the watch and iPhone as remotes to each other to work properly, I think because WCSession issue I mentioned earlier)

erik-apple:
Thanks for this PR Corey! We'll go over it in more detail soon, but at first blush it looks promising.

> It works a little different than the WWDC video as the watch and the iOS are both using ParseCareKit to sync through the cloud remote

Your way is actually better than the way we demonstrated in our WWDC session in some respects. Apple Watch apps can be stand alone apps, or the phone might be out of range, so communicating directly with the server is typically a more robust approach.

> I suspect this is down the line, but adding support for multiple remote will be useful. Specifically with this watchOS case as there are times when a watch and phone are thought to be connected to the Internet, but for some reason don't sync. Allowing them to sync with each other and with a cloud remote could solve the problem. Since the whole knowledge vector is being passed, it seems like could work. From the way it looks like the coordinator works, maybe it would use something similar?

That's an interesting idea!

CareKitStore/CareKitStoreTests/OCKStore/TestStore+BuildRevisions.swift
erik-apple: Sounds right to me!

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
```diff
@@ -266,6 +266,92 @@ class TestStoreSync: XCTestCase {
...
+        let taskUUID = try task.getUUID()
+
+        let outcome = try testStore.addOutcomeAndWait(OCKOutcome(taskUUID: taskUUID, taskOccurrenceIndex: 0, values: [OCKOutcomeValue("test")]))
+        let outcomeUUID = try outcome.getUUID()
+        XCTAssertNoThrow(try testStore.syncAndWait()) //Sync original outcome
+        
+        try testStore.deleteOutcomeAndWait(outcome)
+        XCTAssertNoThrow(try testStore.syncAndWait()) //Sync tombstoned outcome
+        let latestRevisions = dummy2.revisionsPushedInLastSynch
+        XCTAssert(latestRevisions.count == 2)
```
erik-apple, line 286: Wouldn't this ideally be == 1, since we just need to push 1 tombstone?

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
```diff
@@ -266,6 +266,92 @@ class TestStoreSync: XCTestCase {
...
+        let tombstonedOutcomes = latestRevisions.compactMap{
+            entity -> OCKOutcome? in
+            switch entity{
+            case .outcome(let outcome):
+                return outcome
+            default:
+                return nil
+            }
+        }
+        XCTAssert(tombstonedOutcomes.count == 2)
```
erik-apple, line 297: I'm guessing that one of these is the new delete record, and the other is an "update" to the original to tombstone it?

If so, we should see if it's somehow possible to only send the new delete record. I think it should be possible for the client to know what to do given just the new delete record.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
erik-apple: The error seems fine, especially since it's just a unit test.
Consider preferring `.first(where: { try $0.getUUID() == outcomeUUID })` over `.filter({ ... }).first`.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
```diff
@@ -354,6 +440,101 @@ class DummyEndpoint: OCKRemoteSynchronizable {
     }
 }
 
+class DummyEndpoint2: OCKRemoteSynchronizable {
```
erik-apple, line 443: What's the reason for creating a second dummy endpoint class?
Would it be possible to update the existing class to make it do what you want, or to subclass it to create this class?

CareKitStore/CareKitStore/Protocols/CoreData/OCKCoreDataStoreProtocol.swift
```diff
@@ -178,6 +178,7 @@ extension OCKCoreDataStoreProtocol {
             guard let current = currentVersions.first(where: { $0.id == value.id }) else {
                 throw OCKStoreError.invalidValue(reason: "No matching object could be found for id: \(value.id)")
             }
+            current.logicalClock = Int64(context.clockTime)
```
erik-apple, line 181: I'm inferring that this was added to ensure that both the old and new version of the updated object are captured in the next revision. Do you recall why does the old version need to be captured?

CareKitStore/CareKitStore/CoreData/OCKStore+Outcomes.swift
```diff
@@ -151,6 +151,7 @@ extension OCKStore {
         currentOutcomes.forEach {
             $0.deletedDate = Date()
             $0.values = Set()
+            $0.logicalClock = Int64(context.clockTime)
```
erik-apple, line 154: This one does seem necessary. I'd like to come up with a way to get around updating existing objects in a way that requires them to be synchronized a 2nd time, but I think it makes sense to include this, at least for the time being.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: Sure, I'll make this change

CareKitStore/CareKitStore/Protocols/CoreData/OCKCoreDataStoreProtocol.swift
cbaker6: I'm trying to remember, but I believe it had something to do with the old versions "nextVersion" being updated, but not sent to the remote. This ensures that the linkList between versions is updated on the remote side. I think I added a testcase for this also, but will look back through

When this wasn't here, I had to add code on the remote side to repair previous and next version link list because I didn't receive the update from the OCKStore

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: The tombstoned entity is created with a new UUID, along with tombstoning the original outcome with the original UUID. Both are added to revisions

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: Yup, I read this one after posting my previous comment, but this is essentially what I was trying to say about line 286.

I think the problem with knowing what to do on the client side is the change in UUID, it comes with a deleted date, but no reference to the UUID to delete.

If it was like the versioned objects, the tombstone could be the current version, and then the client would know to tombstone the previous version

CareKitStore/CareKitStore/Protocols/CoreData/OCKCoreDataStoreProtocol.swift
erik-apple: That's what I suspected! I think we would prefer to leave it up to the remote to update the previous <--> next link instead of syncing over both the old and new version!

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: The first dummy endpoint was setup to test a specific type of way and I wasn't able to exploit the tombstone issue I was having. The new dummy endpoint is a little bit smarter and when I had it combined with the first one it broke some of the old test cases. I separated them out since they are testing different things. I can look at the possibility of subclassing.

erik-apple: If this one is smarter than the original, maybe we can just replace the original with yours. I haven't looked super closely at the differences yet, but if we can get away with just one dummy class, that would be nice.

Alternatively, if it's too complicated to combine them, maybe we can just give both of them more descriptive names to make it clear what their differences are and which should be used to test what.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: If it tombstoned the same UUID, I agree, it's easy to handle this on the remote

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: Just combined them. The tests pass locally, lets see what happens with travis

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
cbaker6: Fixed

CareKitStore/CareKitStore/Protocols/CoreData/OCKCoreDataStoreProtocol.swift
cbaker6: I can revert, but my intuition is telling me remotes may mess this up since locally a change was made, but never pushed to the server. What I can see happening is if a remote doesn't link properly and a user wants to restore their device from the Cloud then it would be up to the OCKStore to do the same repair type of repair the remote should have done, but on the client side during pullRevisions by checking previousVersion, nextVersion and establishing the link.

CareKitStore/CareKitStore/CoreData/OCKStore+Outcomes.swift
cbaker6: The only way I can think of now is not creating a new tombstoned entity with a different uuid. This goes against the immutability of a record you mentioned before, but I always interpreted this as “most of parts of the record is immutable”. Since fields like the “logicalClock” is never mutated outside of the CareStore (nor is it exposed), it seems okay to me. I guess someone could argue then it’s not really immutable...

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
erik-apple: I think I'm following along now.

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
erik-apple: Thanks!

CareKitStore/CareKitStoreTests/OCKStore/TestStore+Sync.swift
erik-apple: I see where you're coming from. From CareKit's perspective, our diff generation logic is simpler if we don't also send over the tombstone, but I think you would make the argument that it's better to get it right once inside of CareKit than to ask each integration developer to figure out which outcome they need to tombstone.

erik-apple:
Thanks for all the time you've put into this one Corey. If you can get the build passing again, I think this one is ready to merge!

cbaker6:
@erik-apple all of these tests pass locally. I think the random failures have a little to with Travis (slower resources) and Xcode 12 beta (I’m going to take a guess at what’s happening, though I haven’t looked into proof of what I’m about to say...). 

For Xcode 12 beta, when running the unit tests locally on MacBook Pro, it “looks” like the unit tests are executed asynchronously, not synchronously (the tests themselves, not the fetching/saving, etc to the CareKitStore.) I say this because when I run a test, I can see each test quickly spin like they are being executed for a second and this occurs in a round robin way until all of the tests are complete. I’m not sure if Xcode 11 operated in the async manner because I don’t remember seeing the round-robin spinning on each test. Xcode 11 “looked” like it executed tests synchronously. Since the CareKitStore has the largest amount of tests (390) and CareKitTests has the second largest (128), sometimes these tests timeout when running async . When you rerun they pass. If my assumptions above are correct, it may be worthwhile to extend the times to wait for async behavior above 10 seconds (maybe 15 or 20?). This should help the tests complete on slower systems like Travis and maybe even GitHubActions

erik-apple:
I think that's a pretty fair guess I think. Maybe we can bump that up to like 30 seconds.

cbaker6:
@erik-apple the HealthKitLinkage tests looks to be finicky on travis. The latest Xcode travis has is 12 beta 2. I don't see any issues on beta 4 locally. I believe GitHub actions is on beta 4 and they seem to upgrade whenever the latest is released.

erik-apple:
@cbaker6 I pushed a few small changes to match the code style with conventions we follow elsewhere in the codebase. If you're okay with the changes, I'll merge this.

cbaker6:
Looks good to me

PR #454, opened 2020-06-15
Update OCKContactUtility.swift

garyphilipp:
The OCKContactUtility nameFormatter currently uses .medium for the formatter style, which displays given and family names only. Since it is very likely the contacts for Carekit include those in the medical field, i.e., Doctors, it would be more appropriate to use the formatter style .long so that the prefix (Dr.) is included in the formatted string that is displayed.


erik-apple:
Thanks Gary, and congrats on your first contribution to CareKit!

PR #459, opened 2020-06-25
Fix building on watchOS

cbaker6:
The updates are great!

I was playing around with the watchOS example and it wouldn't import CareKit (CareKitStore, CareKitUI, etc. work) via SPM. I added in the fixes that allowed me to get it to build and run


erik-apple:
Thanks! Good catch!

PR #462, opened 2020-06-27
Update travis to xcode12

cbaker6:
It looks like there will be random timeouts from time-to-time due to the resources on travis and probably the time of day. If any of the timeouts become consistent, you can probably increase the wait time on the respective test.

I bumped  `testFetchWithTaskIDsReceivesDeletedTaskNotifications` by a second because it failed once, but I don't think this is a consistent failure. I did it to re-trigger the test.

It should run 5/6 of these concurrently, it's possible the travis settings for the repo is limiting the concurrent builds to 2 (or travis limits the builds for Xcode 12). Since there are more schemes in the updated version, setting the "build latest commit" setting can help canceling builds with updates to PRs

### DIFF
--- a/CKWorkspace.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/CKWorkspace.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -3,7 +3,7 @@
     "pins": [
       {
         "package": "FHIRModels",
-        "repositoryURL": "git@github.com:apple/FHIRModels.git",
+        "repositoryURL": "https://github.com/apple/FHIRModels.git",
         "state": {
           "branch": null,
           "revision": "31dc40c01ef43b89191be5d9719e1ddedb2c91b5",
--- a/CareKitFHIR/CareKitFHIR.xcodeproj/project.pbxproj
+++ b/CareKitFHIR/CareKitFHIR.xcodeproj/project.pbxproj
@@ -22,12 +22,13 @@
 		03AFA931233E69830091DD45 /* OCKR4PatientCoder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03AFA92E233E69830091DD45 /* OCKR4PatientCoder.swift */; };
 		03AFA93D233E74610091DD45 /* FHIRResourceDataTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03AFA93C233E74610091DD45 /* FHIRResourceDataTests.swift */; };
 		03AFA943233E80C40091DD45 /* DSTU2PatientConverterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03AFA942233E80C40091DD45 /* DSTU2PatientConverterTests.swift */; };
-		03C75C5724A50248009F1896 /* ModelsR4 in Frameworks */ = {isa = PBXBuildFile; productRef = 03C75C5624A50248009F1896 /* ModelsR4 */; };
-		03C75C5924A50248009F1896 /* ModelsDSTU2 in Frameworks */ = {isa = PBXBuildFile; productRef = 03C75C5824A50248009F1896 /* ModelsDSTU2 */; };
 		03C85E6E233D71EC002C5901 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03C85E6D233D71EC002C5901 /* Utils.swift */; };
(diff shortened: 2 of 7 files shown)


CKWorkspace.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -3,7 +3,7 @@
     "pins": [
       {
         "package": "FHIRModels",
-        "repositoryURL": "git@github.com:apple/FHIRModels.git",
+        "repositoryURL": "https://github.com/apple/FHIRModels.git",
```
cbaker6, line 6: ssh wouldn't play nicely with SPM in the xcodebuild command on travis, so I switched to https instead. The rest of the changes to the project (except for .travis.yml) are due to this change

erik-apple: I think it makes sense to make this change in the root file too!

CareKitFHIR/CareKitFHIR.xcodeproj/project.pbxproj
```diff
@@ -22,12 +22,13 @@
...
 		03C85E6E233D71EC002C5901 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03C85E6D233D71EC002C5901 /* Utils.swift */; };
 		03C85E73233D7215002C5901 /* CareKitFHIRTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03C85E72233D7215002C5901 /* CareKitFHIRTests.swift */; };
 		03C85E76233D721F002C5901 /* R4PatientConverterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03C85E75233D721F002C5901 /* R4PatientConverterTests.swift */; };
 		03E5929123736D0D00B6E9DE /* DSTU2TaskConverterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03E5929023736D0D00B6E9DE /* DSTU2TaskConverterTests.swift */; };
+		915266B424A7C95E009CB68A /* ModelsDSTU2 in Frameworks */ = {isa = PBXBuildFile; productRef = 915266B324A7C95E009CB68A /* ModelsDSTU2 */; };
+		915266B624A7C95E009CB68A /* ModelsR4 in Frameworks */ = {isa = PBXBuildFile; productRef = 915266B524A7C95E009CB68A /* ModelsR4 */; };
+		915266B824A7C95E009CB68A /* ModelsBuild in Frameworks */ = {isa = PBXBuildFile; productRef = 915266B724A7C95E009CB68A /* ModelsBuild */; };
```
erik-apple, line 31: Did you add the `ModelsBuild` framework because of build errors on Travis, or was this an unintentional change?

CKWorkspace.xcworkspace/xcshareddata/swiftpm/Package.resolved
cbaker6: Updated.

I got these errors while running `swift test` from cl:

```
error: the library 'CareKitFHIR' requires macos 10.10, but depends on the product 'ModelsR4' which requires macos 10.13; consider changing the library 'CareKitFHIR' to require macos 10.13 or later, or the product 'ModelsR4' to require macos 10.10 or earlier.
error: the library 'CareKitFHIR' requires macos 10.10, but depends on the product 'ModelsDSTU2' which requires macos 10.13; consider changing the library 'CareKitFHIR' to require macos 10.13 or later, or the product 'ModelsDSTU2' to require macos 10.10 or earlier.
```

I didn't change the build because I'm not running SPM 5.3 and I'm not sure if automatically resolves this

CareKitFHIR/CareKitFHIR.xcodeproj/project.pbxproj
cbaker6: This was unintentional, as I just selected all of the packages when switching to https. I removed the extra one

CKWorkspace.xcworkspace/xcshareddata/swiftpm/Package.resolved
erik-apple: Hmm, it seems like SPM is trying to run the tests on macOS instead of iOS.

erik-apple:
Thanks for the PR Corey!

PR #464, opened 2020-06-28
Fixed the issue with UILabel's autosizing in high levels of text sizes.

kutakmir:
The issue:

1) Go to settings -> accessibility -> display & text size -> larger text on -> set the size to the second largest or third largest.
2) Run CareKit Catalog target
3) Select Task -> Simple

You'll see a broken layout. This PR fixes this bug.

EDIT: added screenshots.

BEFORE:
Screenshot 2020-06-26 15 02 43

AFTER:
Screenshot 2020-07-03 15 08 56

### DIFF
--- a/CareKitUI/CareKitUI/Common/Labels/OCKLabel.swift
+++ b/CareKitUI/CareKitUI/Common/Labels/OCKLabel.swift
@@ -120,4 +120,10 @@ open class OCKLabel: UILabel, OCKStylable {
     }
 
     open func styleDidChange() {}
+
+    open override func layoutSubviews() {
+        super.layoutSubviews()
+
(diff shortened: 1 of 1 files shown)


gavirawson-apple:
Thanks for the PR and for pointing out this issue @kutakmir! The change you've made definitely fixes the issue, but there seems to be a deeper underlying problem. See #470 for the fix! Since `OCKLabel` is used across many different views, its safest if it is un-opinionated about the `preferredMaxLayoutWidth`.

kutakmir:
@gavirawson-apple you can probably close this PR so that only the PRs that have a potential to be merged are open. Thank you.

PR #477, opened 2020-07-13
Implemented Simple and Detailed Contact Views in SwiftUI

kutakmir:
## Objective
This PR introduces simple and detail contact views in the SwiftUI part of CareKit and CareKitUI.

## Interesting points
In the Detailed Contact View, I couldn't use `.subheadline` Font style because it only allows a single line of text. That's probably a bug in SwiftUI. The same code is being used in the InstructionsTaskView alongside with `.lineLimit(nil)` which doesn't have any effect to the line limit.

The HeaderView didn't contain any image up until now. So included the image. We still don't have a way to populate images for contacts in CareKit so all images are just system placeholders for now.

The size of the image and scale of the fonts doesn't correspond 1:1 to the UIKit version. This is because we are using the predefined sizes. If we wanted to use different sizes for these use-cases, we would have to define them in constants.

Looking forward to your feedback!

Simulator Screen Shot - iPhone 11 - 2020-07-13 at 14 29 49
Simulator Screen Shot - iPhone 11 - 2020-07-13 at 14 29 52
Simulator Screen Shot - iPhone 11 - 2020-07-13 at 14 29 58
Simulator Screen Shot - iPhone 11 - 2020-07-13 at 14 30 00
Simulator Screen Shot - iPhone 11 - 2020-07-13 at 14 43 23


CareKit/CareKit/Shared/Contact/DetailedContactView.swift
```diff
@@ -0,0 +1,138 @@
+//
```
erik-apple, line 1: Nit - remove this first `//` from all headers.

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+    
+    init(controller: OCKDetailedContactController) {
+        contactView = .init(controller: controller) {
+            .init(viewModel: $0.viewModel)
+        }
+    }
+}
+
+// CODE REVIEW: Let's add actions - they are nil right now, see UIKit version
```
erik-apple, line 117: The actions should be closure properties on the `ContactViewModel` that starts a few lines below this.

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    /// The detail text to display in the header.
+    public let detail: String?
+    
+    /// The instructions text to display under the header.
+    public let instructions: String?
+    
+    /// The address text to display under the address button.
+    public let address: String?
+}
```
erik-apple, line 138: You'll want to add some actions into this struct, I believe.

Following the pattern used for the SwiftUI task views, we'll want to add a published contact model property onto the `OCKContactController` class. The controller will be responsible for setting its own methods as the actions on the ContactViewModel.

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
```diff
@@ -0,0 +1,138 @@
...
+        contactView = .init(controller: controller) {
+            .init(viewModel: $0.viewModel)
+        }
+    }
+}
+
+// CODE REVIEW: Let's add actions - they are nil right now, see UIKit version
+private extension CareKitUI.DetailedContactView where Header == _DetailedContactViewHeader, Footer == _DetailedContactViewFooter {
+    init(viewModel: ContactViewModel?) {
+        self.init(title: Text(viewModel?.title ?? ""), detail: Text(viewModel?.detail ?? ""), instructions: Text(viewModel?.instructions ?? ""), image: Image(systemName: "person.crop.circle"), disclosureImage: nil, callButton: ContactButton(title: Text("Call"), image: Image(systemName: "phone"), action: nil), messageButton: ContactButton(title: Text("Message"), image: Image(systemName: "text.bubble"), action: nil), emailButton:  ContactButton(title: Text("E-mail"), image: Image(systemName: "envelope"), action: nil), addressButton: AddressButton(title: Text("Address"), detail: Text(viewModel?.address ?? ""), image: Image(systemName: "location"), action: nil))
```
erik-apple, line 120: This line is really long. Consider putting in line breaks after each argument.

CareKit/CareKit/Shared/Extensions/OCKContactController+Extensions.swift
```diff
@@ -0,0 +1,69 @@
...
+    var viewModel: ContactViewModel? { makeViewModel(from: contact) }
+    
+    private func makeViewModel(from contact: OCKAnyContact?) -> ContactViewModel? {
+        guard let contact = contact else { return  nil }
+
+        return .init(title: formatName(contact.name), detail: contact.title, instructions: contact.role, address: formatAddress(contact.address))
+    }
+
+    // CODE REVIEW: Why is the formatName parameter optinal?
+    private func formatName(_ name: PersonNameComponents?) -> String {
```
erik-apple, line 47: It's optional because the view model might be empty if it hasn't finished loading yet. When `OCKStore`, queries are basically instant, but sometimes when people use other stores or load things from a server, views can appear on screen before their data is ready.

We need to be able to produce a valid string to display even if the model isn't available yet, so we make name optional to account for that case.

CareKit/CareKit/Shared/Extensions/OCKContactController+Extensions.swift
```diff
@@ -0,0 +1,69 @@
...
+            return ""
+        }
+        let formatter = PersonNameComponentsFormatter()
+        formatter.style = .medium
+        
+        return formatter.string(from: name)
+    }
+    
+    // CODE REVIEW: Why is the address parameter optinal?
+    // CODE REVIEW: Why is the return type non-optional?
```
erik-apple, line 58: See my comment above. The parameter is optional because the data might not be ready yet, but the return type is non-optional because we need to display something no matter what.

CareKit/CareKitTests/Contact/TestDetailedContactView.swift
```diff
@@ -0,0 +1,63 @@
...
+class TestDetailedContactView: XCTestCase {
+
+    let controller: OCKDetailedContactController = {
+        let store = OCKStore(name: "carekit-store", type: .onDisk)
+        return .init(storeManager: .init(wrapping: store))
+    }()
+
+    let query = OCKContactQuery(id: "lexi-torres")
+    let data = OCKContact(id: "lexi-torres", givenName: "Lexi", familyName: "Torres", carePlanUUID: nil)
+    // CODE REVIEW: Let's break this long line down into multiple lines
```
erik-apple, line 49: Yes please!

kutakmir:
@erik-apple thanks for the feedback and explanation on some of the things - I'll follow up shortly!

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
kutakmir: Done

CareKit/CareKitTests/Contact/TestDetailedContactView.swift
kutakmir: done

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
kutakmir: done

CareKit/CareKit/Shared/Extensions/OCKContactController+Extensions.swift
kutakmir: That makes sense. Thanks for clarifying!

CareKit/CareKit/Shared/Contact/DetailedContactView.swift
kutakmir: done

Issue #481, opened 2020-07-16
SwiftUI support for ChecklistTaskView


paulofaria:
Hello, I'm creating this issue mainly to find out if there is work being done about porting ChecklistTaskView to SwiftUI. If there's no one working on this I'll gladly assign this issue to myself and create a PR implementing it.

kutakmir:
Hi @paulofaria, I have this already in progress, I sent you an email.

PR #482, opened 2020-07-16
Add ChecklistTaskView

paulofaria:
Todos:

[ ] Add support for images and detail disclosure in the header.
[ ] Fine-tune UI so it matches UIKit as close as possible.
[ ] Add tests.

Solves #481.

### DIFF
--- a/CareKit/CareKit/iOS/Task/ChecklistTaskView.swift
+++ b/CareKit/CareKit/iOS/Task/ChecklistTaskView.swift
@@ -1,0 +1,216 @@
+/*
+ Copyright (c) 2020, Apple Inc. All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+
+ 1.  Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+
+ 2.  Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following disclaimer in the documentation and/or
+ other materials provided with the distribution.
+
+ 3. Neither the name of the copyright holder(s) nor the names of any contributors
+ may be used to endorse or promote products derived from this software without
+ specific prior written permission. No license is granted to the trademarks of
+ the copyright holders even if such marks are included in this software.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#if !os(watchOS)
+
+import CareKitStore
+import CareKitUI
+import Foundation
+import SwiftUI
+
+/// A card that displays a header, instructions, and indicators for progress and goal values.
+///
+/// In CareKit, this view is intended to display a particular event for a task.
+///
+/// # View Updates
+/// The view updates with the observed controller. By default, data from the controller is mapped to the view. The mapping can be customized by
+/// providing a closure that returns a view. The closure is called whenever the controller changes.
+///
+/// # Style
+/// The card supports styling using `careKitStyle(_:)`.
+///
+/// ```
+///     +-------------------------------------------------------+
+///     | +------+                                              |
+///     | | icon | [title]                       [detail        |
+///     | | img  | [detail]                       disclosure]   |
+///     | +------+                                              |
+///     |                                                       |
+///     |  --------------------------------------------------   |
+///     |                                                       |
+///     |  +-------------------------------------------------+  |
+///     |  | [title]                                   [img] |  |
+///     |  +-------------------------------------------------+  |
+///     |  +-------------------------------------------------+  |
+///     |  | [title]                                   [img] |  |
+///     |  +-------------------------------------------------+  |
+///     |                         .                             |
+///     |                         .                             |
+///     |                         .                             |
+///     |  +-------------------------------------------------+  |
+///     |  | [title]                                   [img] |  |
+///     |  +-------------------------------------------------+  |
+///     |                                                       |
+///     |  [instructions]                                       |
+///     +-------------------------------------------------------+
+/// ```
+@available(iOS 14.0, *)
+public struct ChecklistTaskView<Header: View, Content: View>: View {
+
+    private typealias TaskView = SynchronizedTaskView<OCKChecklistTaskController, CareKitUI.ChecklistTaskView<Header, Content>>
+
+    private let taskView: TaskView
+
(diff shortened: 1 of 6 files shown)


erik-apple:
Thanks for the PR!
This looks like a great start. 

I like your proposal for the `ChecklistItemIdentifiable` protocol.
This seems like a principal that could be applied in more than just the checklist view, and since its only requirement is an `isComplete` boolean, it might be worth renaming it to something like `Completable` or `Toggleable` to make it semantically recyclable for other use cases.

paulofaria:
Yeah, I thought about that as well, nice! I prefer `Completable` over `Toggleable`. Where should I move it in the file hierarchy?

CareKitUI/CareKitUI/iOS/Task/ChecklistTaskView.swift
```diff
@@ -0,0 +1,307 @@
...
+}
+
+public extension ChecklistTaskView {
+    /// Create an instance.
+    /// - Parameters:
+    ///   - instructions: Instructions text to display under the header.
+    ///   - items: Items to display in the checklist.
+    ///   - action: Action to perform when a checklist item is tapped.
+    ///   - header: Header to inject at the top of the card. Specified content will be stacked vertically.
+    init<Item : Completable, ChecklistContent : View>(
```
gavirawson-apple, line 163: The ability to customize the checklist content is great! 

Currently the view is generic over the `Content`, which allows a developer to replace the entire checklist. When creating the view with custom `Content`, the initializer and layout will look very similar to some other views we have in framework already (see `InstructionsTaskView` when providing a custom `Header` and `Footer`). 

I think it's safe to tie this view to the checklist style, and only allow for customization of rows in the checklist. What do you think? The structure can look something like this:

```swift
public struct ChecklistTaskView<Header: View, ItemContent: View> {

    init<Item: Identifiable>(
        instructions: Text? = nil,
        items: [Item],
        @ViewBuilder header: () -> Header,
        @ViewBuilder itemContent: (Item) -> ItemContent
    ) {
        ...
    }
}

public struct _ChecklistTaskViewItemContent: View {
    ...
}

public struct ChecklistTaskViewItem: Identifiable {
    var id: String
    var title: String
    var isComplete: Bool
}

public extension ChecklistTaskView where Header == HeaderView {

    init<Item: Identifiable>(
        title: Text,
        detail: Text? = nil,
        instructions: Text? = nil,
        items: [Item],
        @ViewBuilder itemContent: (Item) -> ItemContent
    ) {
        ...
    }
}

public extension ChecklistTaskView where ItemContent == _ChecklistTaskViewItemContent {

    init(
        instructions: Text? = nil,
        items: [ChecklistTaskViewItem],
        @ViewBuilder header: () -> Header,
    ) {
        ...
    }
}

public extension ChecklistTaskView where Header == HeaderView, ItemContent == _ChecklistTaskViewItemContent {

    init(
        title: Text,
        detail: Text? = nil,
        instructions: Text? = nil,
        items: [ChecklistTaskViewItem],
    ) {
        ...
    }
}
```

CareKitUI/CareKitUI/Shared/Completable.swift
```diff
@@ -0,0 +1,36 @@
...
+ FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/// Protocol used to express entities that can be completed.
+public protocol Completable : Identifiable {
```
gavirawson-apple, line 33: If you decide to go with the architecture I mentioned above, this protocol won't be needed anymore.

CareKit/CareKit/iOS/Task/ChecklistTaskView.swift
```diff
@@ -0,0 +1,216 @@
...
+
+    init(viewModel: ChecklistTaskViewModel?) {
+        self.init(
+            title: Text(viewModel?.title ?? ""),
+            detail: viewModel?.detail.map(Text.init),
+            items: viewModel?.items ?? [],
+            action: viewModel?.action ?? { _ in },
+            instructions: viewModel?.instructions.map(Text.init)
+        ) { item in
+            Text(item.title)
```
gavirawson-apple, line 181: This initializer should delegate to an initializer on the view in CareKitUI that builds the default view using the view model. It's nice to have an initializer like that, in case the developer just wants the view as it comes out of the box. I think as things are set up right now, there is no initializer for the default view - the initializer for the default view will not provide any View Builders parameters.

If you decide to go with the architecture I mention below, the initializer will look like:

```swift
init(
    title: Text,
    detail: Text? = nil,
    instructions: Text? = nil,
    items: [ChecklistTaskViewItem],
) { 
    ....
}
```

CareKitUI/CareKitUI/iOS/Task/ChecklistTaskView.swift
paulofaria: Hey, @gavirawson-apple. Thank you, those are great suggestions. You mentioned something I honestly found a bit odd with the current architecture. IMO, the current SwiftUI views are _too_ generic with _too_ specific names. I think it would be good to have some generic views and then wrap them with the more specific views. For example, the following code:

```swift
InstructionsTaskView(header: { Text("Header") }, footer: { Text("Footer") })
SimpleTaskView(header: { Text("Header") }, detailDisclosure: { Text("Detail Disclosure") })
NumericProgressTaskView(header: { Text("Header") }, content: { Text("Content") })
LabeledValueTaskView(header: { Text("Header") }, detailDisclosure: { Text("Detail Disclosure") })
```

Produces the following views:

Capture d’écran 2020-07-21 à 08 29 05

As you can see `InstructionsTaskView` and `NumericProgressTaskView` behave identically and its view contents don't have any connection with their type names. Same thing for `SimpleTaskView` and `LabeledValueTaskView`. I think the best path would be to create some rudimentary types, maybe something like `HCardView` and `VCardView`, then build the more specific types with more specific initializers on top of these rudimentary types?

paulofaria: I also see version 2.1 hasn't been released yet. I really think this should be addressed before we release 2.1.

paulofaria: If you, @gavirawson-apple and @erik-apple, think this makes sense. I can create another PR around this.

paulofaria: Thinking more about it, I believe we don't even need something like `HCardView` or `VCardView`, since it would be super trivial to replicate its behavior. I think we should just make the task views behave more specifically. Exposing `CardView` with `HeaderView` already provides most of the flexibility users need.

gavirawson-apple: That's a really good point that we have thought about on our end as well. It definitely seems odd. Ultimately, I think the answer lies in your and the community's hands. Let's take the `InstructionsTaskView` as an example. Does exposing just the `HeaderView` and `CardView` allow for enough customization points?

I can offer our thought process that ultimately led to the decision of creating more generic containers like the `Footer` in the `InstructionsTaskView`:

- When designing the API for a view, we try to think about the most likely spots a developer will want to customize. There's a very wide range of use cases to which CareKit is being applied, so we try to lean toward a customizable and open API.

- Say a developer wants to customize the completion button in the `InstructionsTaskView`. Maybe their branding dictates a certain button style that is different from the one provided. If we do not expose the customizable `Footer`, they will have to recreate the view in CareKitUI themself. If they would like a synchronized view as well, they would have to create an associated view in CareKit. That can lead to a lot of extra code and burden on the developer for something as simple as swapping out the button style.

- Currently the `InstructionsTaskView` and the `NumericProgressTaskView` look identical in their most generic form.  However, it's possible that one day in the future the `NumericProgressTaskView ` may take on new UI elements that differentiates it from the `InstructionsTaskView `. And those UI elements might not be customizable through the public API.  If that happens, it will probably be a bit clearer why the `Content` in the `NumericProgressTaskView` is customizable. For example, maybe the view takes on some progress bars in the footer that demonstrate some weekly and monthly statistics for a user's progress towards a goal, for more of a bird's eye view of progress. With the API architected as it is today, we can make that change to the view without breaking the API. 

Those are a few reasons that led us to the architecture we have today. But I'm interested in hearing your thoughts! All feedback is super appreciated and helps us learn how developers are consuming the framework, ultimately leading to a better API!

PR #492, opened 2020-08-04
Use GitHub Actions for CI

cbaker6:
GitHub Actions runs faster and runs 5 builds in parallel. I think the times will be a lot better than travis along with it being more consistent with failures.

I don't think it's showing here yet because the actions file needs to be merged first, but you can view how the [PR](https://github.com/cbaker6/CareKit/pull/6) looks on my local branch. The results are [here](https://github.com/cbaker6/CareKit/actions).


erik-apple:
This seems like a nice thing to have. I saw on your fork that it's building all 6 targets. Is it also running the unit tests for them?

cbaker6:
Yup it’s running the unit tests on each. It will also handle `swift build/test” well once that’s fixed.

erik-apple:
What do you think of a test file like this?
It's a bit more terse and reduced duplication of values that we often change, namely the version of iOS being targeted.
```yml
name: CI

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]


jobs:
        
  test:
    
    runs-on: macos-latest
  
    strategy:
      matrix: 
        destination: ['platform=iOS\ Simulator,OS=14.0,name=iPhone\ 11\ Pro\ Max']
        scheme: ['CareKit\ iOS', 'CareKitStore\ iOS', 'CareKitUI\ iOS', 'CareKitFHIR', 'OCKSample', 'OCKCatalog']
  
    steps:
      - uses: actions/checkout@v2
      - name: Set Xcode Version
        run: sudo xcode-select -s /Applications/Xcode_12.app
      - name: Build
        run: set -o pipefail && env NSUnbufferedIO=YES xcodebuild -verbose -workspace CKWorkspace.xcworkspace -scheme ${{ matrix.scheme }} -destination ${{ matrix.destination }} build test | xcpretty
```

erik-apple:
Great, can you swap that into this PR?

cbaker6:
The one thing I did in the previous version was remove `OS=14.0`, I believe it uses the default for Xcode. I did this so the yml didn't need to be updated every time the Xcode version in actions updates.

It's currently in the yml, what would you like me to do with it?

erik-apple:
We actually prefer that it be explicit because it can sometimes prevent CI from breaking unexpectedly!

If tests break because CI begins using a different version of the OS without us being aware, it can take a while to realize what changed. If builder changes are always the direct result of a PR and source change, it's much easier to pin down the root cause if version bumps lead to breaks.

cbaker6:
Not sure what was up with GitHub Actions and why it was failing when this was merged, but re-running should do the trick on the master branch. I reran it on my fork and all is well

erik-apple:
I'm not sure why it's failing now either. It's failing on #499 too, and rerunning it isn't doing the trick. I can't reproduce the problem locally either.

cbaker6:
Weird as I just merged the latest update to my master and it worked. I also merged the latest to the slider view and it works there on the PR.

One thing I see is that there's a bug somewhere in the condensed yml because one 1 test fails, it fails all of the others

PR #499, opened 2020-08-28
Note->Task cardinality should be 0..1

drdavec:
Note to Task relationship cardinality is 0..* and should be 0..1


erik-apple:
Thanks for catching this!
That's definitely a mistake.

PR #508, opened 2020-10-01
OCKOutcomeValue requiring all key values

cbaker6:
It appears the problem I was having was deeper than my original fix, but I found the issue.

OCKOutcomeValue is currently encoding all key/value pairs, even when nil, and requiring all be present when decoding, but they can't be nil at that time. This leads to an OCKOutcomeValue requiring non-nil values for all key/values during decoding.

I believe this issue is only in OCKOutcomeValue because it's the only one that has a custom encoder/decoder, which seems to be there for `value`. The issue is `encode` and `decode` are used instead of `encodeIfPresent` and `decodeIfPresent`. I believe the latter is used by default by the memberwise encoder/decoder. I didn't apply "IfPresent" to `value` and `timezone` since those are required. I'm guessing `String?.self` and the like were there to work as `encodeIfPresent`, but it looks like `?` doesn't stop it from throwing an error, not even with `try?`.

As an example, if you are encoding from and `OCKOutcomeValue` that had `kind = nil`, and then you attempt to decode it:

```swift
let encoded = try JSONEncoder().encode(self)
try JSONDecoder().decode(OCKOutcomeValue.self, from: encoded)
```

It will throw an error.

Update: The custom encoder/decoder was missing `asset` and `notes`. I added those as well


cbaker6:
As far as the tests not passing, I didn't change anything with tasks, but I do see the same tests failing locally.

It appears to be a problem with querying for DateInterval, it seems to only be able to query for the newer task as oppose to finding an older one

cbaker6:
Another observation on the "random" occurrence of test case failures. They seem to be the same 2 tests that fail and were failing on my computer ~midnight, when I run them this morning, they pass. This may have something to do with the tests taking the current date and using `addingTimeInterval`. Maybe there's something weird when the date interval crosses over from midnight or different days? This may explain the random failures at different times of the day.

CareKitStore/CareKitStoreTests/Structs/TestOutcomeValue.swift
```diff
@@ -116,6 +116,78 @@ class TestOutcomeValue: XCTestCase {
             testEqualityOfEncodings(outcome1: value1, outcome2: value2)
         }
     }
+    
+    func testProperDecodingWhenMissingValues() throws {
+        let valueToDecode = "{\"value\": 10,\"timezone\": {\"identifier\": \"America/New_York\"},\"type\": \"\(OCKOutcomeValueType.integer.rawValue)\"}"
+
+        guard let data = valueToDecode.data(using: .utf8) else{
```
erik-apple, line 123:
```suggestion
        guard let data = valueToDecode.data(using: .utf8) else {
```

erik-apple (changes requested):
Thanks for being so thorough! I'll look into those failing tests as well. In other places I've been setting dates to a fixed time (usually 0:00). Hopefully that'll fix the issue here too. Let me know if you figure it out in the meantime!

cbaker6:
@erik-apple If I'm remembering correctly, the other time the random test failures pop up are around ~12pm. If so, and you are able to replicate the failure on your system, you might be able to verify there's no bug in the query filter (I think some of the other test cases verify there's no issue here). I didn't think to check this last night...

erik-apple:
I haven't been able to replicate those test failures locally yet. If they ever fail for you locally, please record the exact time so we can try to replicate it on our end too!

cbaker6:
Verified the two tests don't work on the main branch.

For `testProperDecodingWhenMissingValues`, the following error is thrown: "Default: failed: caught error: "keyNotFound(CodingKeys(stringValue: "kind", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"kind\", intValue: nil) (\"kind\").", underlyingError: nil))". Of course, if you add an entry for `kind`, the error will propagate to the next missing key/value.

For `testProperDecodingWhenMissingValues`, the test cases fail for missing `asset` and `note` in the decoded `OCKOutcomeValue`

cbaker6:
> I haven't been able to replicate those test failures locally yet. If they ever fail for you locally, please record the exact time so we can try to replicate it on our end too!

It was failing for me locally, consistently ~12:30am - 1:00am EST. No matter how many times I ran it, the same test cases that fail in CI failed on my computer.

erik-apple:
This looks good to me!

erik-apple (approved):

PR #522, opened 2020-11-25
Only display current version of contacts

### LINKED ISSUE #523: OCKContactListViewController showing complete history of contacts
When using the OCKContactListViewController to display the list of contacts, it will show the complete history of all contacts added, even if they have a different version whether it is an updated version of the contact, or the contact has been deleted.

When working with querying for contacts and fetching them, I was able to see what was happening with the OCKContactListViewController. Not specifying a date in the query leads to fetching all the same contacts that are displayed in the OCKContactListViewController, however, when I do specify a date in the query (specifically for the current date) the fetched contacts are current whether it is updated contacts or being that the deleted contacts no longer show.

Querying works fine, but the OCKContactListViewController is what I believe is querying contacts wrong.

cbaker6:
Currently, the `OCKContactsListViewController` queries and shows all versions (even deleted ones) of contacts. Passing in date makes it only show active versions.

This is related to [version section](https://github.com/carekit-apple/CareKit#store-) of the README.

This bug was discovered by a @ChrisGambrell, an undergraduate in my iOS App Development course.

Fix #523


gavirawson-apple (approved):

gavirawson-apple:
Thanks for the fix @cbaker6 !

Issue #535, opened 2020-12-27
Cannot add OCKHealthKitTask to OCKCarePlan


drdavec:
I am trying to add a new OCKHealthKitTask to an existing OCKCarePlan, but the stored health kit task is always saved and fetched with a null carePlanUUID.  Tracing this through, it appears to be a problem in OCKCoreDataTaskStore.swift, in func copyTask.

On this line, the fetchObject returns nil, even tho the planUUID is valid.  This works for non-CareKit tasks added to the same planUUID.
        if let planUUID = task.carePlanUUID { persistableTask.carePlan = try? fetchObject(uuid: planUUID) }

I suspect that fetchObject is attempting to find a the planUUID within a OCKHealthKitTaskStore, which is not the correct store.  There also does not appear to be any way to create an OCKCarePlan within a OCKHealthKitTaskStore.

I can't see any workaround for this.  Any insight into this error or a possible solution?

Thanks!
  Dave Carlson

erik-apple:
Dave, thanks for bringing this to our attention. Yes, this is currently a limitation of HealthKit linked CareKit tasks. We'll need to look into a way to support your use case.

erik-apple:
@drdavec Thanks for raising this issue. It was fixed as part of our latest PR.

When you create an instance of `OCKHealthKitPassthroughStore`, you're now required to pass a reference an existing `OCKStore`. Doing so will allow us to make the proper associations.

PR #553, opened 2021-02-19
Synchronization Improvements

erik-apple:
This PR contains a series of changes that improve the robustness of CareKit's synchronization algorithm.

Synchronization now happens in two phases. The first phase, merging, is now provably conflict free since CareKit's store can be shown to be a CRDT at the data structure level. The second phase, discrepancy resolution, is where logically concurrent changes are resolved. These changes to the underlying synchronization mechanism make the implementation considerably simpler.

There are several noteworthy API changes landing as part of this PR.
1. Versioned objects `next` and `previous` properties are now arrays of UUIDs instead of optional UUIDs.
2. OCKOutcome is now a versioned object
3. Properties like tags, notes, remoteID and others have been removed from notes, schedules, and outcome values. The same values are available on each type's parent entity.
4. The `OCKAny___Query` protocols have been removed. Use the corresponding concrete query types instead.
5. The `uuid` associated with each CareKit object is no longer optional
6. Signatures for `OCKRemoteSynchronizable` protocol methods now reflect that there may be N-way conflicts instead of just 2-way conflicts
7. `OCKHealthKitPassthroughStore` is now initialized with a reference to an existing `OCKStore`


srinathtm-apple (approved):

erik-apple:
This PR also closes out issue #535.

PR #563, opened 2021-03-26
Fix #562

### LINKED ISSUE #562: OCKDailyTasksPageViewController tasksDelegate behavior
I noticed that there is a OCKDailyTasksPageViewControllerDelegate available on OCKDailyTasksPageViewController but no real way to attach to it since it is assigned to self in viewDidLoad.

https://github.com/carekit-apple/CareKit/blob/730037b1a3dd9bb18d61e8a2d0a88bb022ab963c/CareKit/CareKit/iOS/Higher%20Order/ViewController/OCKDailyTasksPageViewController.swift#L63-L66

Furthermore, in the following code, the nil coalescing operator doesn't have any effect since self is assigned to tasksDelegate already.

https://github.com/carekit-apple/CareKit/blob/730037b1a3dd9bb18d61e8a2d0a88bb022ab963c/CareKit/CareKit/iOS/Higher%20Order/ViewController/OCKDailyTasksPageViewController.swift#L115-L117

Perhaps tasksDelegate should be assigned to self in init() rather than viewDidLoad()?

erik-apple:
The daily tasks page view controller should not set itself as its own delegate. Instead, it should use its own implementation of delegate methods if there is no delegate set.

@nizar Can you checkout this branch and confirm that it fixes #562 before we merge it?

### DIFF
--- a/CareKit/CareKit/iOS/Higher Order/ViewController/OCKDailyTasksPageViewController.swift
+++ b/CareKit/CareKit/iOS/Higher Order/ViewController/OCKDailyTasksPageViewController.swift
@@ -50,20 +50,14 @@ public protocol OCKDailyTasksPageViewControllerDelegate: OCKTaskViewControllerDe
 
 /// Displays a calendar page view controller in the header, and a collection of tasks
 /// in the body. The tasks are automatically queried based on the selection in the calendar.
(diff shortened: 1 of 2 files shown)


srinathtm-apple (approved):

gavirawson-apple (approved):

nizar:
@erik-apple LGTM!

PR #565, opened 2021-03-29
Only return one version of an outcome when a date interval is given

erik-apple:
This PR is a proposed fix for #558.

When a date interval is specified on an outcome query, only one version of the outcome should be returned. Previously we were returning several, in error.

### DIFF
--- a/CareKitStore/CareKitStore/CoreData/OCKStore+Outcomes.swift
+++ b/CareKitStore/CareKitStore/CoreData/OCKStore+Outcomes.swift
@@ -117,9 +117,29 @@ extension OCKStore {
         var predicate = query.basicPredicate(enforceDateInterval: false)
         
         if let interval = query.dateInterval {
-            let beforePredicate = NSPredicate(format: "%K < %@", #keyPath(OCKCDOutcome.startDate), interval.end as NSDate)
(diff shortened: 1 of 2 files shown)


cbaker6:
I can try this out in ParseCareKit real quick and let you know if I see the issue I saw before.

erik-apple:
That'd be great, thanks!

There was definitely a bug lurking in here, so we'll want to get this fix in even if it turns out not to fully resolve the issue we were investigating when it was discovered.

cbaker6:
np, I'll provide an update in a few minutes

srinathtm-apple (approved):

cbaker6:
So when I log into the sample account I created a few days ago a few times (3) to view the data I used to test before. The data from a few days ago (nausea using the log button) consistently shows correctly.

The data I added today isn't showing correctly, but let me verify I didn't do something wrong.

cbaker6:
I think it may be working. I have to fix a bug, but if you merge, I'll report back after my bug fix

cbaker6:
Update, I fixed my bug and I suspect this PR fixes the issue @nizar reported as I was able to replicate the same problem through syncing

erik-apple:
Great, thanks for helping verify this Corey!

PR #574, opened 2021-04-20
Corrected OCKTask documentation

ZachJBurns:
The documentation is wrong here in OCKTask.swift

Fixed instance of 'OCKCarePlan' where it should be 'OCKTask'
Corrected id and title parameter description.
Removed healthKitLinkage since it does not belong to the initialization.


erik-apple (approved):

erik-apple:
Thanks for submitting a fix!

PR #579, opened 2021-05-05
Use asset when available in detail card view

### LINKED ISSUE #524: Issue adding an image to an OCKChecklistTaskViewController card
I cannot find a way to add an image to an OCKChecklistTaskViewController card. In the attachment, the top portion above the instructions and label is light gray- this is where the image _should_ go.

github issue carekit 2

cbaker6:
If a card is displaying an `OCKAnyTask` it will automatically display the asset tied to the task when details are tapped. Close #524 

@erik-apple and @gavirawson-apple I couldn't find testcases where I should add tests for this. Let me know where to look and I can add them if needed

### DIFF
--- a/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
+++ b/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
@@ -1,4 +1,3 @@
-//
 /*
  Copyright (c) 2021, Apple Inc. All rights reserved.
  
--- a/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
+++ b/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
@@ -1,5 +1,6 @@
+//
 /*
- Copyright (c) 2019, Apple Inc. All rights reserved.
+ Copyright (c) 2021, Apple Inc. All rights reserved.
  
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
@@ -27,44 +28,18 @@
  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-#if canImport(UIKit)
 
-import Contacts
-import MapKit
+import Foundation
 import UIKit
 
-struct OCKContactUtility {
-    private static let addressFormatter: CNPostalAddressFormatter = {
-        let formatter = CNPostalAddressFormatter()
-        formatter.style = .mailingAddress
-        return formatter
-    }()
-
-    private static let nameFormatter: PersonNameComponentsFormatter = {
-        let nameFormatter = PersonNameComponentsFormatter()
-        nameFormatter.style = .long
-        return nameFormatter
-    }()
-
-    static func string(from location: CNPostalAddress?) -> String? {
-        guard let location = location else { return nil }
-        return addressFormatter.string(from: location)
-    }
-
-    static func string(from components: PersonNameComponents?) -> String? {
-        guard let components = components else { return nil }
-        return nameFormatter.string(from: components)
-    }
-/*
-    static func image(from asset: String?) -> UIImage? {
-        guard let asset = asset else { return nil }
-
+extension UIImage {
+    static func asset(_ name: String?) -> UIImage?{
         // We can't be sure if the image they provide is in the assets folder, in the bundle, or in a directory.
+        guard let asset = name else { return nil }
         // We can check all 3 possibilities and then choose whichever is non-nil.
         let symbol = UIImage(systemName: asset)
         let appAssetsImage = UIImage(named: asset)
         let otherUrlImage = UIImage(contentsOfFile: asset)
         return otherUrlImage ?? appAssetsImage ?? symbol
-    }*/
+    }
 }
-#endif
--- a/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
+++ b/CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
@@ -35,11 +35,11 @@ import UIKit
 extension UIImage {
     static func asset(_ name: String?) -> UIImage?{
         // We can't be sure if the image they provide is in the assets folder, in the bundle, or in a directory.
-        guard let asset = name else { return nil }
+        guard let name = name else { return nil }
         // We can check all 3 possibilities and then choose whichever is non-nil.
-        let symbol = UIImage(systemName: asset)
-        let appAssetsImage = UIImage(named: asset)
-        let otherUrlImage = UIImage(contentsOfFile: asset)
+        let symbol = UIImage(systemName: name)
+        let appAssetsImage = UIImage(named: name)
+        let otherUrlImage = UIImage(contentsOfFile: name)
         return otherUrlImage ?? appAssetsImage ?? symbol
     }
 }
--- a/CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
+++ b/CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
@@ -444,7 +444,7 @@ open class OCKTaskController: ObservableObject {
         let detailViewController = OCKDetailViewController(showsCloseButton: true)
         detailViewController.detailView.titleLabel.text = task.title
         detailViewController.detailView.bodyLabel.text = task.instructions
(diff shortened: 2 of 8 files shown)


erik-apple:
I wonder if perhaps a better approach here would be to include an optional `asset` property on `OCKAnyTask`. Then no type casting would be required and we'd still get the desired behavior.

cbaker6:
> I wonder if perhaps a better approach here would be to include an optional `asset` property on `OCKAnyTask`. Then no type casting would be required and we'd still get the desired behavior.

I was going to do this, but I didn't see asset in OCKAnyTask, let me look again...

cbaker6:
It looks like asset isn't part of `OCKAnyTask` from what I can tell.

cbaker6:
Looking further back, it looks like the only OCKAny to have asset was OCKAnyContact and the other entities gained it through conforming to the OCKObjectCompatable protocol

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
```diff
@@ -444,6 +444,9 @@ open class OCKTaskController: ObservableObject {
         let detailViewController = OCKDetailViewController(showsCloseButton: true)
         detailViewController.detailView.titleLabel.text = task.title
         detailViewController.detailView.bodyLabel.text = task.instructions
+        if let asset = task.asset {
+            detailViewController.detailView.imageView.image = UIImage(named: asset)
+        }
```
erik-apple, line 449:
```suggestion
        detailViewController.detailView.imageView.image = task.asset.map(UIImage.init(named:))        
```

CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
```diff
@@ -55,6 +55,9 @@ public protocol OCKAnyTask {
     /// An universally unique identifier for this task.
     var uuid: UUID { get }
 
+    /// A string representation of an asset.
+    var asset: String? { get }
+
```
erik-apple, line 60: 👍

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
erik-apple: This one-liner might be just a little bit cleaner. Check to make sure it compiles though!

cbaker6: It looks like it returns `UImage??`.

cbaker6:
The last commit assumes `UImage(named: "")` will always will return nil which I don't know if that's true

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
cbaker6: `detailViewController.detailView.imageView.image = task.asset.map(UIImage.init(named:)) ?? nil` would work. Let me know which of the options compared to below works best

cbaker6:
Let me know if you want me to remove the AppIcon asset from being used for this in the sample app. I wanted to provide an example without adding additional assets.

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
```diff
@@ -444,6 +444,7 @@ open class OCKTaskController: ObservableObject {
         let detailViewController = OCKDetailViewController(showsCloseButton: true)
         detailViewController.detailView.titleLabel.text = task.title
         detailViewController.detailView.bodyLabel.text = task.instructions
+        detailViewController.detailView.imageView.image = task.asset.map(UIImage.init(named:)) ?? nil
```
erik-apple, line 447: Looking at this again, I think we need to do something just a bit more nuanced. It's not guaranteed that the asset string is the name of an image in the assets catalog. It could also be an SFSymbol or name of a file in the app bundle.

There is a static method on `OCKContactUtility` that might be useful here.

```swift
    static func image(from asset: String?) -> UIImage? {
        guard let asset = asset else { return nil }

        // We can't be sure if the image they provide is in the assets folder, in the bundle, or in a directory.
        // We can check all 3 possibilities and then choose whichever is non-nil.
        let symbol = UIImage(systemName: asset)
        let appAssetsImage = UIImage(named: asset)
        let otherUrlImage = UIImage(contentsOfFile: asset)
        return otherUrlImage ?? appAssetsImage ?? symbol
    }
```

erik-apple (changes requested):

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
cbaker6: Sounds good to me. Let me see if I can add it

cbaker6: That uses optional binding for asset, should I go back to it?

erik-apple: The reason that method takes an optional is so that you can write things like `image(from: task.asset)` cleanly on one line!

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
```diff
@@ -444,7 +444,13 @@ open class OCKTaskController: ObservableObject {
...
         detailViewController.detailView.titleLabel.text = task.title
         detailViewController.detailView.bodyLabel.text = task.instructions
-        detailViewController.detailView.imageView.image = task.asset.map(UIImage.init(named:)) ?? nil
+        // We can't be sure if the image they provide is in the assets folder, in the bundle, or in a directory.
+        guard let asset = task.asset else { return detailViewController }
+        // We can check all 3 possibilities and then choose whichever is non-nil.
+        let symbol = UIImage(systemName: asset)
+        let appAssetsImage = UIImage(named: asset)
+        let otherUrlImage = UIImage(contentsOfFile: asset)
+        detailViewController.detailView.imageView.image = otherUrlImage ?? appAssetsImage ?? symbol
```
erik-apple, line 453: I think we should extract this into a method on something in CareKit. We'll probably want to use it in other places as well. Perhaps it could live in a `UIImage+Asset.swift` file in the extensions directory.

```swift
extension UIImage {
    static func asset(_ name: String?) -> UIImage? { /*...*/ }
}
```

erik-apple (changes requested):

OCKSample/OCKSample/AppDelegate.swift
```diff
@@ -85,7 +85,7 @@ private extension OCKStore {
         var doxylamine = OCKTask(id: "doxylamine", title: "Take Doxylamine",
                                  carePlanUUID: nil, schedule: schedule)
         doxylamine.instructions = "Take 25mg of doxylamine when you experience nausea."
-        doxylamine.asset = "AppIcon"
+        doxylamine.asset = "pills"
```
erik-apple, line 88: Great idea!

CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
cbaker6: I removed the example you showed in `OCKContactUtility` and replace the usage with the extension.

CareKit/CareKit/iOS/Utilities/OCKContactUtility.swift
```diff
@@ -55,7 +55,7 @@ struct OCKContactUtility {
         guard let components = components else { return nil }
         return nameFormatter.string(from: components)
     }
-
+/*
```
erik-apple, line 58: It's okay to delete this instead of just commenting it out!

CareKit/CareKit/Shared/Extensions/UIImage+Asset.swift
```diff
@@ -0,0 +1,45 @@
+//
```
erik-apple, line 1: Nit: delete this `//` on the first line so that it matches all the other files.

erik-apple (changes requested):

CareKit/CareKit/iOS/Utilities/OCKContactUtility.swift
cbaker6: Removed

erik-apple (approved):

erik-apple:
Great work Corey! LGTM!

erik-apple:
I like that this will add an example of how to use SFSymbols to the sample app!

cbaker6:
> I like that this will add an example of how to use SFSymbols to the sample app!

I was hoping to find some SF symbols to match some of the other tasks, but didn't see any other good matches.

erik-apple:
Thanks for the PR Corey! 🎉

PR #610, opened 2021-07-02
Removes call to objectWillSend within Published properties of OCKTaskController subclasses

### LINKED ISSUE #608: Question - usage of objectWillSend within Published property on ObservableObject
Hi there.  In `OCKSimpleTaskController` there appears to be usage of `objectWillSend` despite the property being marked `Published`.  (refer to [main:OCKSimpleTaskController.swift](https://github.com/carekit-apple/CareKit/blob/5c3b14a4c2076e00c235c0357c490493908d1eb8/CareKit/CareKit/Shared/Task/Controller/OCKSimpleTaskController.swift#L38))

For convenience:
```swift
    @Published public private(set) var viewModel: SimpleTaskViewModel? {
        willSet { objectWillChange.send() }
    }
```

This seems superfluous (would it cause duplicate notifications?), but I may be missing something - is this intentional and serve a purpose? 

(If its incorrect/superfluous, I'd be more than happy to push a correction for review.)

Thanks - love the framework - looking forward to contributing!

rstrahl:
Prior to iOS 14.5, if a subclass of an `ObservableObject` has published properties, modifying their values will not automatically publish changes through the `objectWillChangePublisher`, forcing developers to call `objectWillChange.send()` within a `willSet` observer. 

This incorrect behaviour was resolved in iOS 14.5 ([release notes](https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-14_5-release-notes)):

> Using Published in a subclass of a type conforming to ObservableObject now correctly publishes changes. (71816443)

Closes #608

### DIFF
--- a/CareKit/CareKit/Shared/Task/Controller/OCKInstructionsTaskController.swift
+++ b/CareKit/CareKit/Shared/Task/Controller/OCKInstructionsTaskController.swift
@@ -34,9 +34,7 @@ import Foundation
 open class OCKInstructionsTaskController: OCKTaskController {
 
     /// Data used to create a `CareKitUI.InstructionsTaskView`.
-    @Published public private(set) var viewModel: InstructionsTaskViewModel? {
-        willSet { objectWillChange.send() }
-    }
+    @Published public private(set) var viewModel: InstructionsTaskViewModel?
 
     private var cancellable: AnyCancellable?
 
--- a/CareKit/CareKit/Shared/Task/Controller/OCKSimpleTaskController.swift
+++ b/CareKit/CareKit/Shared/Task/Controller/OCKSimpleTaskController.swift
@@ -34,9 +34,7 @@ import Foundation
 open class OCKSimpleTaskController: OCKTaskController {
 
     /// Data used to create a `CareKitUI.SimpleTaskView`.
-    @Published public private(set) var viewModel: SimpleTaskViewModel? {
-        willSet { objectWillChange.send() }
-    }
+    @Published public private(set) var viewModel: SimpleTaskViewModel?
 
     private var cancellable: AnyCancellable?
 


gavirawson-apple (approved):
Thanks for the change!

gavirawson-apple:
I'm thinking more about your thought about considering backwards compatibility. While this change won't affect the public API, I think it will change the behavior of the framework for developers using an earlier iOS version. Do you think that rather than deleting the `willSet` block, we can wrap the logic in the block in an availability mark to help avoid that issue?

rstrahl:
I think thats a good call, yeah - I'll make that change.  

Looks like I also need to investigate a failed test in `CareKitStore` - it surprises me because there wasn't a fail when executing locally.

gavirawson-apple:
I believe that failing test is a known issue with CI, no need to worry about it.

rstrahl:
In revising to include the `#available` block around the `objectWillChange.send()` call, I've realized that it would effectively cause a regression in the behaviour; the `willSet` block would still fire, but it doesn't inherit the superclass' correct behaviour - it overrides it.

I'm now skeptical of the value of this pull request; its original objective was to remove code and simplify, and I don't see how we can achieve that while supporting iOS versions < 14.5 (as we should).

gavirawson-apple:
You bring up a good point, that's definitely possible. In that case, as you suggested, I would recommend closing out this PR. There is currently work being done on our SwiftUI API that will change the shape of the controller and make this problem a thing of the past. Thanks for spotting the issue for us!

rstrahl:
@gavirawson-apple I'm really interested in learning more about those changes; is there a branch somewhere that I can take a look? I'm prototyping an app currently, and have been interested in helping contribute to evolve the SwiftUI support.

rstrahl:
Closing PR as per comments above.

gavirawson-apple:
We do not currently have a branch up, as the work is in the earlier stages. But, I think that might be a really nice idea in the future - we'd very much appreciate your contributions in that branch.

PR #611, opened 2021-07-04
Modify access modifiers in OCKTaskController and OCKTaskControllerError to support external subclasses and extensions

rstrahl:
`OCKTaskController` is marked as an open class, and when attempting to follow the guidance in the README for subclassing it for use with contextual task views (e.g. `OCKInstructionsTaskView`), or writing extensions to support use-cases, there are some limitations when attempting to do this externally from the CareKit framework itself.  This pull request makes the following modifications to support extension outside of the framework:

* Removes the `internal` access modifier from the setter on the `error` property, making it public.
* Adds the `public` access modifier to both `validatedViewModel()` and `validatedEvent(forIndexPath:)` methods to enable external subclasses/extensions to leverage those when creating new methods for manipulating events and outcomes, using a similar pattern to those established by methods like `setEvent()` or `appendOutcomeValue()`.
* Adds the `public` access modifier to `OCKTaskControllerError` to support the changes in (2) above, as well as enable external subclasses/extensions to leverage that type in other code that may throw those errors.

### DIFF
--- a/CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
+++ b/CareKit/CareKit/Shared/Task/Controller/OCKTaskController.swift
@@ -43,7 +43,7 @@ open class OCKTaskController: ObservableObject {
     @Published public final var taskEvents = OCKTaskEvents()
 
     /// The error encountered by the controller.
-    @Published public internal(set) var error: Error?
+    @Published public var error: Error?
 
     /// The store manager against which the task will be synchronized.
     public let storeManager: OCKSynchronizedStoreManager
@@ -497,14 +497,14 @@ private extension AnyCancellable {
     }
 }
 
-enum OCKTaskControllerError: Error, LocalizedError {
+public enum OCKTaskControllerError: Error, LocalizedError {
 
     case emptyTaskEvents
     case invalidIndexPath(_ indexPath: IndexPath)
     case noOutcomeValueForEvent(_ event: OCKAnyEvent, _ index: Int)
     case cannotMakeOutcomeFor(_ event: OCKAnyEvent)
 
-    var errorDescription: String? {
+    public var errorDescription: String? {
         switch self {
         case .emptyTaskEvents: return "Task events is empty"
         case let .noOutcomeValueForEvent(event, index): return "Event has no outcome value at index \(index): \(event)"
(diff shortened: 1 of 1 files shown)


gavirawson-apple:
Thanks for the PR @rstrahl! These types/properties are implementation details that are likely to change, especially as we continue to explore a SwiftUI API. It may be wise to leave them as internal so that developers don't begin to depend on them before they are removed or changed.

Most of these are slim implementations and can be recreated in your specific project. Take the error for example - you can subclass the task controller and add your own custom error property that you can use as needed. 

I am curious about how you intend to use the error. Since the error comes from the store, the issue generally comes from the developer and not the user. Therefore, it might not be necessary to surface an error at all, and we might be able to use assertions or logging. Would you agree with that, or are there cases where you need the user to have access to the error in some way (I.E an alert)?

rstrahl:
The use-case you describe re: subclassing the task controller is exactly the direction I was headed in - the subclass is in my project, rather than in a fork, so the `internal` modifiers on the error property and type prohibit me from having my subclass re-use those.  I can definitely move forward by recreating those lightweight classes, like you've described - this PR was an effort keep things simple and re-usable, and determine whether such a change aligned with the direction the SwiftUI API is headed in.

Let me know if I should hold off on things like this, and I can close the PR - eager to see the upcoming updates!

gavirawson-apple:
That all makes a lot of sense, opening up these types will definitely save time writing code.

If there's a need to handle some sort of error, in your subclass you can create a new error property without having to rely on the internal one:

```swift
class CustomTaskController: SimpleTaskController {

    // You can create your own custom error publisher and handle errors as needed.
    // Can optionally use the more recognized delegate pattern here as well.
    let errorPublisher: AnyPublisher<Error, Never>
    
    ....
}
```

There is one tricky piece here that isn't handled. Interactions such as button presses always succeed in the UI to ensure the UI feels responsive. A good example is the circular completion toggle in the SimpleTaskView - If an error occurs when that button is toggled (IE an outcome couldn't be created/deleted), the UI needs to be reloaded to reflect the actual data in the database. If you're encountering that scenario in your work, perhaps:

1. We expose some function named `refresh()` or `didEncounterError()`. That method will reset the view, a tricky "gotcha" that isn't immediately apparent to developers.

    In the custom controller described above, that would look like:

    ```swift
    errorPublisher.sink {
       refresh()
    }
    ```

2. We avoid exposing the properties proposed in this PR. They definitely help to avoid writing more code, but the code seems straightforward enough to write and we'll avoid confusion if they change in a future release.

What are your thoughts on that proposal?

rstrahl:
I think that makes sense, and it helps establish some API guidance in differentiating between between errors in the fundamental CareKit logic and third-party/custom classes. 

I'll close out this PR - as always, thanks for your help and insight!

gavirawson-apple:
No problem, really appreciate your hard work and contributions here! It helps us understand what API hooks are useful!

PR #617, opened 2021-08-05
Async/Await extensions for OCKStoreProtocol Constituents

JWStaiert:
I have added async/await extensions to all constituent protocols of OCKStoreProtocol and fixed miscellaneous documentation typos in those protocols.

These changes compile cleanly with Xcode 13 beta 4 and do run on iOS 15 simulator. However I haven't tested every method, nor have I committed any of my unit tests. If anyone feels unit tests are needed, please comment.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -21,6 +21,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set Xcode Version
-        run: sudo xcode-select -s /Applications/Xcode_13.app
+        run: sudo xcode-select -s /Applications/Xcode_12.4.app
       - name: Build
         run: set -o pipefail && env NSUnbufferedIO=YES xcodebuild -verbose -workspace CKWorkspace.xcworkspace -scheme ${{ matrix.scheme }} -destination ${{ matrix.destination }} build test | xcpretty
--- a/CareKitStore/CareKitStore/Protocols/CarePlans/OCKAnyCarePlanStore.swift
+++ b/CareKitStore/CareKitStore/Protocols/CarePlans/OCKAnyCarePlanStore.swift
@@ -106,7 +106,7 @@ public protocol OCKAnyCarePlanStore: OCKAnyReadOnlyCarePlanStore {
     /// `deleteAnyCarePlan` asynchronously deletes a single care plan from the store.
     ///
     /// - Parameters:
-    ///   - plans: n single care plan to be deleted. The care plans must exist in the store.
+    ///   - plans: A single care plan to be deleted. The care plans must exist in the store.
     ///   - callbackQueue: The queue that the completion closure should be called on. In most cases this should be the main queue.
     ///   - completion: A callback that will fire on the provided callback queue.
     func deleteAnyCarePlan(_ plan: OCKAnyCarePlan, callbackQueue: DispatchQueue, completion: OCKResultClosure<OCKAnyCarePlan>?)
@@ -146,7 +146,6 @@ public extension OCKAnyCarePlanStore {
     }
 }
 
-// MARK: JW
 // MARK: Async methods for OCKAnyReadOnlyCarePlanStore
 
 @available(iOS 15.0, *)
@@ -175,7 +174,6 @@ public extension OCKAnyReadOnlyCarePlanStore {
     }
 }
 
-// MARK: JW
 // MARK: Async methods for OCKAnyCarePlanStore
 
 @available(iOS 15.0, *)
--- a/CareKitStore/CareKitStore/Protocols/CarePlans/OCKAnyCarePlanStore.swift
+++ b/CareKitStore/CareKitStore/Protocols/CarePlans/OCKAnyCarePlanStore.swift
@@ -37,7 +37,7 @@ public protocol OCKAnyReadOnlyCarePlanStore: OCKAnyResettableStore {
     /// In `CareKit` apps, the delegate will be set automatically, and it should not be modified.
     var carePlanDelegate: OCKCarePlanStoreDelegate? { get set }
 
-    /// `fetchCarePlans` asynchronously retrieves an array of care plans from the store.
+    /// `fetchAnyCarePlans` asynchronously retrieves an array of care plans from the store.
     ///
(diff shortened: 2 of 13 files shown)


erik-apple (changes requested):
Thanks for your hard work on this Jason!

These changes look good to me. I just requested two small changes. There's one typo, and we can also remove the `// MARK: JW` comments.

erik-apple:
I think it's fine that there aren't unit tests for each of these methods. They're just convenience wrappers around methods that are already thoroughly unit tested.

Perhaps in the future we could update the existing unit tests to use these new methods instead. That would get us a little bit better coverage without introducing any new tests, and we could remove the `OCKStoreProtocol+Synchronous.swift` file altogether!

JWStaiert:
Changes made.

JWStaiert:
Am I also responsible for updating the workflow?

CI is failing because the build is using Xcode 12. Package.swift may also need updating.

erik-apple:
> Am I responsible for updating ".github/workflows/swift.yml"?

If it needs to be updated, it should be done as part of this PR. I can make the change for you if you allow administrators to make changes to your branch, or you can do it yourself.

What needs to be updated in `Package.swift`?

JWStaiert:
I'll do it. And will verify if Package.swift needs and update (to the platform iOS version.)

cbaker6:
GitHub actions doesn’t support the Xcode beta publicly. To get it to build for actions as well is for anyone else who doesn’t have beta the added code/files should be wrapped in.

```swift
#if swift(>=5.5)
 import Foundation

 @available(iOS 15.0, watchOS 15.0, *)

#endif
```

.github/workflows/swift.yml
```diff
@@ -21,6 +21,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set Xcode Version
-        run: sudo xcode-select -s /Applications/Xcode_12.app
```
cbaker6, line 24: You will need to switch this back to Xcode 12, 13 isn't available.

Current software: https://github.com/actions/virtual-environments/blob/macOS-10.15/20210801.1/images/macos/macos-10.15-Readme.md

You can check the issues in the actions repo for info on Xcode 13

cbaker6: If you select 12.4, you will need to update the simulator version:

https://github.com/carekit-apple/CareKit/blob/59146eb13b3812db9872ce9b7c8574562b6aba9e/.github/workflows/swift.yml#L18

You should confirm with @erik-apple about these updates

JWStaiert:
Erik,

No rush, sometime after the weekend is great: when you have a moment would you comment on @cbaker6 's request?

.github/workflows/swift.yml
cbaker6: I recommend switching this back to Xcode 12 as 12.4 as those changes won't enable the build of your PR (async/await). The reason I recommend this is say for example, there are other reasons the CareKit team has not to jump to a higher version of Xcode, you don't want that getting in the way of your PR being accepted.

 I opened #618 to allow the CareKit team assess the upgrade to Xcode 12.4 separately from this PR.

erik-apple (changes requested):
This is looking good!

I have one more request before we merge this work in. Could you open a new issue on GitHub to remove the `#if swift(>=5.5)` bits once Xcode 13 is supported by CI?  Then, copy a link to that issue and paste it into the source code right above the places where we're using conditional compilation? That should help us remember why those compiler directives are there and remind us to remove them once it becomes possible to.

```swift
// Remove this once Xcode 13 is available on GitHub actions
// https://github.com/carekit-apple/carekit/issues/???
#if swift(>=5.5)
```

.github/workflows/swift.yml
```diff
@@ -21,6 +21,6 @@ jobs:
...
+        run: sudo xcode-select -s /Applications/Xcode_13.app
       - name: Build
```
erik-apple, line 25: You'll need to either need to remove this commit so that it doesn't conflict with the changes from Corey that we just landed, or change this line to 12.4, so that it doesn't overwrite Corey's change.

cbaker6:
> This is looking good!
> 
> I have one more request before we merge this work in. Could you open a new issue on GitHub to remove the `#if swift(>=5.5)` bits once Xcode 13 is supported by CI? Then, copy a link to that issue and paste it into the source code right above the places where we're using conditional compilation? That should help us remember why those compiler directives are there and remind us to remove them once it becomes possible to.
> 
> ```swift
> // Remove this once Xcode 13 is available on GitHub actions
> // https://github.com/carekit-apple/carekit/issues/???
> #if swift(>=5.5)
> ```

If you remove this, the builds after the removal won't work with iOS 13 and watchOS older than 9. From what I've seen, older Xcode versions will acknowledge `#if swift(>=5.5)`, but will try to build `@available(iOS 15.0, watchOS 9.0, *)` even if those OS's are not available. If the goal is to only have CareKit for iOS 15+ then of course what I mentioned doesn't matter.

erik-apple:
There is an important distinction here between Xcode 13 and iOS 15.

Once Xcode 13 is released, using `@available(iOS 15.0, watchOS 9.0, *)` will compile even for iOS 13 and 14. 

This is not true for Xcode 12, because Xcode 12 doesn't know what iOS 15 and watchOS 9 are, so the compiler directives are required.

All that to say that CareKit will eventually require Xcode 13, but will still build for iOS 14.

cbaker6:
> There is an important distinction here between Xcode 13 and iOS 15.
> 
> Once Xcode 13 is released, using `@available(iOS 15.0, watchOS 9.0, *)` will compile even for iOS 13 and 14.
> 
> This is not true for Xcode 12, because Xcode 12 doesn't know what iOS 15 and watchOS 9 are, so the compiler directives are required.
> 
> All that to say that CareKit will eventually require Xcode 13, but will still build for iOS 14.

Makes sense, thanks for the clarification!

erik-apple (approved):
Awesome! Thanks for getting this across the finish line Jason!

erik-apple:
Congrats on your first CareKit PR!
I know a lot of people will appreciate this one!

PR #618, opened 2021-08-08
Bump CI to Xcode 12.4

cbaker6:
Xcode 12.4 is the latest publicly available in Actions. It looks like the CI has been running agains 12.0 for awhile. Current software: https://github.com/actions/virtual-environments/blob/macOS-10.15/20210801.1/images/macos/macos-10.15-Readme.md

This will need to be bumped manually from time-to-time because of the discussion here: https://github.com/carekit-apple/CareKit/pull/492#issuecomment-675167708

Maybe this gets rid of CI failures since it's not running on older Xcode? Or at least make them less frequent.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -15,12 +15,12 @@ jobs:
   
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


erik-apple (approved):

erik-apple:
I'm happy to take this change! Thanks for applying the update!

PR #639, opened 2022-01-29
add async/await support to iOS 13

cbaker6:


erik-apple (commented):
This looks great! I’ll look over this soon! Thank you!

cbaker6:
@erik-apple I added some missing properties to `OCKHealthKitOutcome`. Though some of these aren't used currently, I made them match `OCKOutcome`.

I also marked some important properties as `internal` in the concrete struct. I did this because I believe they were internal before the winter update and they probably shouldn't be manipulated by developers. Let me know if you need me to make some adjustments.

YanceeMcDaniel (approved):

PR #649, opened 2022-03-31
Update OCKChartController.swift

ikonted:
I would like to have the option to have a custom event query.


gavirawson-apple (commented):
This looks great @ikonted! Thanks for the contribution, I think the flexibility of creating a custom event query will make this controller much more helpful.

Since these changes expose a property and an initializer to the public API, would you like to try your hand at documenting the two for future clients? There is some other documentation in this class that can be used as a reference / starting point.

Issue #665, opened 2022-08-09
Xcode 14/Ventura Beta5 issue with compiling carekitstore


wmalloc:
<img width="1398" alt="Screenshot 2022-08-08 at 9 51 40 PM" src="https://user-images.githubusercontent.com/125501/183567347-94bebd53-fef4-4758-b0b2-bd4058577815.png">

gavirawson-apple:
I'm not seeing this issue on my end. Can you share your configuration (MacOS version, Xcode version, and CareKit branch)?

Of note - it also looks the CareKit SPM package is being built here, is that intentional?

cbaker6:
When building a project that uses CareKit (latest commit on main) via SPM on macOS Ventura, the CareKitStore produces the following error:

<img width="1083" alt="image" src="https://user-images.githubusercontent.com/8621344/200882342-685d4fb6-7636-4d16-958f-ed37c2469d7d.png">

This error doesn't occur when building on macOS Monterey using Xcode 14.1.

Though apps can still build and run with this error, Xcode Cloud will not build successfully because of it. To build a successful archive, the build has to be completed manually.

zxcheergo:
And How to fix this error? Should we wait for a newer version of CareKit?

gavirawson-apple:
Thanks for the additional info folks, I'm going to look into this on our end.

rachelnaidich:
I had the same issue, and it seemed to go away when I cloned the package and used it locally instead.

Hengyu:
Hi team, do we have any updates on this issue? This does not block the build & compile for local tests but does block the archive process which prevents us from releasing new versions. @gavirawson-apple

cbaker6:
@gavirawson-apple I see this error when using the latest issue which seems to be closer to the problem:

> Attempt to add read-only file at path file:///Users/xxxx/Documents/Xcode_Projects/ParseCareKit/DerivedData/ParseCareKit/SourcePackages/checkouts/CareKit/CareKitStore/CareKitStore/CoreData/Migrations/2_0To2_1/2.0_2.1_Mapping.xcmappingmodel/xcmapping.xml read/write. Adding it read-only instead. This will be a hard error in the future; you must specify the NSReadOnlyPersistentStoreOption.

gavirawson-apple:
Yep I see that on my end too, thanks Corey.

Hengyu:
Thanks @gavirawson-apple, l'll try the fallback instructions.

PR #679, opened 2023-01-13
Fix Xcode errors when using Framework in SPM

cbaker6:
One way to partially address #665 is to exclude the migrations file, `CoreData/Migrations/2_0To2_1/2.0_2.1_Mapping.xcmappingmodel` when using SPM. Though this is not ideal, we can recommend that users who need to migrate from an old version of OCKStore use drop the CareKit project into their projects directly as specified in https://github.com/carekit-apple/CareKit/issues/665#issuecomment-1377562666.

People already using the current data store won't be effected by this change.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -40,10 +40,7 @@ let package = Package(
         .target(
             name: "CareKitStore",
             path: "CareKitStore/CareKitStore",
-            exclude: ["Info.plist"],
-            resources: [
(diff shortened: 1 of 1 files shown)


gavirawson-apple:
I'd like to dig into the place where we use the mapping model before we think about removing it. I wouldn't want this to disappear if the framework specifies the file during the migration process. If migration is left up to the client, removing it might be okay for now.

PR #689, opened 2023-04-06
3.0 Release

gavirawson-apple:
## CareKit 3.0 

CareKit 3.0 introduces new tools to integrate care data into your SwiftUI views, including tools that stream care data and that aggregate progress across care tasks. With the help of these tools, let CareKit handle persistence of user data while you focus on building out engaging visual experiences for your users.

We’d love to hear what you think of the new updates. Feel free to leave a general note or idea here, or open up a new issue if you have any specific questions.

Happy coding!

### Release Notes

*CareKit 3.0* supports *iOS* and *watchOS* and requires *Xcode 14.0* or later. The minimum supported *Base SDK* is *13.0* for *iOS* and *6.0* for *watchOS*.

*CareKit 3.0* includes the following new features and enhancements by Apple Inc. (https://github.com/carekit-apple)

#### SwiftUI Support

- Adds a SwiftUI state property that retrieves and stays synchronized with data in a care store (`CareStoreFetchRequest`).

- Adds the ability to pass a care store through the SwiftUI environment. Accessing the store in the environment without first setting it will trigger a runtime crash.

- Adds new types for results fetched from a care store (`CareStoreFetchedResults`, `CareStoreFetchedResult`).

- Adds convenience initializers for CareKitUI views using data fetched from a care store (`InstructionsTaskView`, `SimpleTaskView`, `LabeledValueTaskView`, `NumericProgressTaskView`) 

#### UIKit View Controllers

- Initializers that previously required a synchronized store manager now require a store that conforms to `OCKAnyStoreProtocol` instead.

- Removes the concept of a controller from the view controllers that display care data. The responsibilities previously handled by the controller are now handled by the view controllers and the streams outputting live data from the care store. 

- Adds initialization parameters to a few view controllers that allow for specifying a task progress computation strategy for a particular task (EX: `OCKWeekCalendarPageViewController`).  

- Individual view synchronizers have been replaced with a single protocol (`OCKViewSynchronizing`) with requirements for synchronizing a UIKit view with data fetched from the care store. 
 
- Initializers that previously accepted an adherence aggregator now require a closure that's used to compute task progress instead. 

- View controllers that display data from a care store now subclass `SynchronizedViewController`. This class can be subclassed to create a custom view controller that is synchronized with data in a care store. 

#### Store

- Adds the ability to stream data from a care store using Swift Concurrency (`CareStoreQueryResults`). Data can either be streamed from Core Data (`OCKStore`) or HealthKit (`OCKHealthKitPassthroughStore`).

[...]

PR #555, opened 2021-02-21
Fix OCKStore inits

cbaker6:
The inits for `OCKCDHealKitLinkage`, `OCKCDHealthKitTask`, and `OCKCDPostalAddress` are causing app crashes. The other entities seem to be calling self instead of super, so I switched these to self and didn't see crashes anymore.

Fix #554

PR #691, opened 2023-04-07
Fix version 3.0 beta build

cbaker6:
- [x] Fix all tests to run on latest version of Xcode provided by GitHub Actions (Xcode 14.2 on macOS 12)
- [x] Make `store` in `OCKDailyPageViewController.swift` public readable, commit: https://github.com/carekit-apple/CareKit/pull/691/commits/38be8cee49697292e7b08a4c50e7fa67151cd5d5
- [x] Fix  app crash caused by `OCKCDPostalAddress` init, commit: https://github.com/carekit-apple/CareKit/pull/691/commits/863a5e8b3d224c23bff78929e01ae0f5027dcccf
- [x] Minor doc nits like: `careKitStore->careStore` commit: https://github.com/carekit-apple/CareKit/pull/691/commits/4bf2d81ea8acb54b6428a26dca796ceda5f05153 
- [x] Fix broken watchOS build. There are currently some files that shouldn't build for watchOS
- [x] Run CI on `iOS` and `watchOS` to prevent breaking PR's on `watchOS` 
- [x] Align all `IPHONEOS_DEPLOYMENT_TARGET` to `14`
- [x] Align all `WATCHOS_DEPLOYMENT_TARGET ` to  `7.0`
- [x] Fix `CareKitFHIR` project to build/test on `watchOS` 
- [x] Fix broken testcases 
- [x] Make any new commits cancel previous running actions on PR
- [x] Bump `AsyncAlgorithms` to `0.1.0`
- [x] Add .spi.yml for properly building on [Swift Package Index](https://swiftpackageindex.com/carekit-apple/CareKit ). Swift Package Index is the successor of [SwiftPM Library](https://swiftpackageindex.com/images/swiftpm-library.png) and [sponsored by Apple](https://blog.swiftpackageindex.com/posts/welcoming-apple-as-a-supporter-of-the-swift-package-index/)
- [x] Let SPI build docs as well. SPI will automatically build docs from the main branch along with any regular/beta release allowing all versions of docs to be captured and visited at anytime  
- [x] Add SPI badges:
image

Note1: The watchOS platform should appear after this PR is merged to the main as the added `.spi.yml` tells Swift Package Index what scheme to use for building watch.


cbaker6:
@gavirawson-apple This is ready for review, let me know if you want me to make any adjustments.

cbaker6:
@gavirawson-apple It looks like there are a couple files missing that SPM can see, but prevents testing. These files are currently not part of the Xcode build which is why it succeeds:

- `TestCoreDataQueryPublisher.swift` - this is in the repo, but not in the Xcode project. Currently can't build with it because it's missing the types below
- `CoreDataSource`, `QueryPublisher`

CareKitStore/CareKitStoreTests/Structs/TestKnowledgeVector.swift
```diff
@@ -35,7 +35,7 @@ import XCTest
 class TestKnowledgeVector: XCTestCase {
 
     func testEncoding() throws {
-        let id = UIDevice.current.identifierForVendor!
```
cbaker6, line 38: UIDevice isn't available on watchOS

cbaker6:
I've also tested these changes locally in Xcode 14.3 and fix all errors/warnings.

CareKit/CareKitTests/TestAnyEventStoreExtensions.swift
```diff
@@ -118,6 +118,10 @@ class TestAnyEventStoreExtensions: XCTestCase {
             didUpdate.fulfill()
         }
 
+        #if compiler(>=5.8.0)
+        await fulfillment(of: [didUpdate], timeout: 2)
+        #else
         wait(for: [didUpdate], timeout: 2)
```
cbaker6, line 124: This is a warning in Swift 5.8.0 that will become an error in Swift 6.0+

SunburstEnzo: Good catch with this, a similar change is needed in CareKitStore/TestCoreDataQueryMonitor.swift in testIrrelevantChangeDoesNotProduceResult. I do wonder if `await withCheckedContinuation` might be more suitable but that'll require significant changes to both tests.
Screenshot 2023-04-08 at 11 27 55 pm

CareKit/CareKit/iOS/Calendar/Paging/OCKWeekCalendarPageViewController.swift
```diff
@@ -83,7 +83,8 @@ UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelega
     /// The initial date displayed when the view controller is loaded.
     private let startingDate = Date()
 
-    private let store: OCKAnyStoreProtocol
+    /// The store the view controller uses for synchronization.
+    public let store: OCKAnyStoreProtocol
```
cbaker6, line 87: Made this store along with the one in `OCKDailyPageViewController` public readable. The reason for this is when a developer subclasses they may want to fetch/save data to the relevant `store` in the view. If it's not public, the developer will need to access a store by other means (maybe global?) which can actually be a different store than the one in the view.

Public access allows for something similar to the `careStore` environment, but in UIKit, and preservers the behavior when OCKSynchronizedStoreManager was public readable.

CareKitFHIR/CareKitFHIR.xcodeproj/project.pbxproj
```diff
@@ -430,7 +430,9 @@
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
 				ONLY_ACTIVE_ARCH = YES;
-				SDKROOT = iphoneos;
+				SDKROOT = auto;
```
cbaker6, line 433: These types of changes are similar across all `xcodeproj` files and allow the projects and respective targets to build for iPhone and watchOS

CareKitStore/CareKitStore/CoreData/OCKCDPostalAddress.swift
```diff
@@ -41,8 +41,8 @@ class OCKCDPostalAddress: NSManagedObject {
     @NSManaged var country: String
     @NSManaged var isoCountryCode: String
     
-    init(address: OCKPostalAddress, context: NSManagedObjectContext) {
-        super.init(entity: Self.entity(), insertInto: context)
+    convenience init(address: OCKPostalAddress, context: NSManagedObjectContext) {
```
cbaker6, line 44: This was causing an app crash, similar fix was done previously in #555

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.6
+// swift-tools-version:5.7
```
cbaker6, line 1: Tested the build and the lowest it will build on is `5.7` which reflects Xcode 14.0 mentioned in the 3.0.0 release

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
```diff
@@ -88,7 +88,7 @@ public struct OCKScheduleElement: Codable, Equatable {
         }
     }
 
-    private let calendar: Calendar
+    private var calendar = Calendar.current
```
cbaker6, line 91: #689 adds `calendar` to be a required encoded/decoded property which will break any `OCKTask` that's currently stored on a server in JSON because previous versions never contained this property.

Making the property a `var` and setting a default value will allow older saved OCKTasks's to be decoded as well as new ones (preserving all new updates with backwards compatibility). The other way would be to write a custom decoder for `OCKSheduleElement`, but that's more work and I don't think it's necessary.

CareKit/CareKitTests/TestAnyEventStoreExtensions.swift
cbaker6: I don't believe that will work as `wait` is known to cause deadlocks in XCTest and `fulfillment` was introduced to solve the issue. Some details are in [Xcode 14.3 changelog](https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes)

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
```diff
@@ -132,6 +130,7 @@ public struct OCKScheduleElement: Codable, Equatable {
     public init(start: Date, end: Date?, interval: DateComponents, text: String? = nil,
                 targetValues: [OCKOutcomeValue] = [], duration: Duration = .hours(1)) {
 
+        let calendar = Calendar.current
```
cbaker6, line 133: #689 adds `calendar` to be a required encoded/decoded property which will break any `OCKTask` that's currently stored on a server in JSON because previous versions of CareKit never contained this property.

~Making the property a `var` and setting a default value will allow older saved OCKTasks's to be decoded as well as new ones (preserving all new updates with backwards compatibility).~ The other way would be to write a custom decoder for `OCKSheduleElement`, but that's more work and I don't think it's necessary.

Decoding Calendar.current after a few minutes seemed to cause crashes after ~10 minutes when `makeValue` is called during the hydration of a OCKTask from CoreData. Checking to see if this latest commit fixes the crash.

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
```diff
@@ -88,8 +88,6 @@ public struct OCKScheduleElement: Codable, Equatable {
         }
     }
 
-    private let calendar: Calendar
```
cbaker6, line 91: #689 adds `calendar` to be a required encoded/decoded property which will break any `OCKTask` that's currently stored on a server in JSON because previous versions of CareKit never contained this property.

~Making the property a `var` and setting a default value will allow older saved OCKTasks's to be decoded as well as new ones (preserving all new updates with backwards compatibility).~ The other way would be to write a custom decoder for `OCKSheduleElement`, but that's more work and I don't think it's necessary.

Decoding Calendar.current after a few minutes seemed to cause crashes after ~10 minutes when `makeValue` is called during the hydration of a OCKTask from CoreData. Checking to see if this latest commit fixes the crash.

CareKit/CareKit/iOS/Contact/View Controllers/OCKContactViewResponder.swift
```diff
@@ -159,6 +159,16 @@ final class OCKContactViewResponder: NSObject,
             action: #selector(dismissViewController)
         )
         
+        /*
+         TODO: Remove in the future. Explicitly setting the tint color here to support
+         current developers that have a SwiftUI lifecycle app and wrap this view
+         controller in a `UIViewControllerRepresentable` implementation...Tint color
+         is not propagated...etc.
+         */
+        let tintColor = contactViewController.determineTintColor(from: contactView)
```
cbaker6, line 168: All of the tintColor changes are simply the updates I added in #676, but in the correct places for 3.0.0

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
gavirawson-apple: Nice catch. I wonder if a good solution here is to pass in the calendar when it's needed and store it elsewhere (maybe the OCKStore?).

> OCKScheduleElement after a few minutes seems to cause crashes after ~10 minutes when makeValue is called during the hydration of a OCKTask from CoreData.

Can you expand on that a bit? Curious what you discovered and if the issue is caused by this stored calendar?

CareKit.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
gavirawson-apple, line 1: I think file is supposed to be ignored by git?

CareKit/CareKit.xcodeproj/project.pbxproj
gavirawson-apple, line 1: Can you walk me through the changes in this file and why they were needed?

OCKCatalog/OCKCatalog.xcodeproj/project.xcworkspace/contents.xcworkspacedata
gavirawson-apple, line 1: Should this file be ignored too?

.spi.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+builder:
+   configs:
+   - documentation_targets: ["CareKit", "CareKitUI", "CareKitStore", "CareKitFHIR"]
```
gavirawson-apple, line 4: Interesting, where do the built docs end up?

.spi.yml
```diff
@@ -0,0 +1,9 @@
...
+     platform: ios
+   - platform: ios
+     scheme: "CareKit"
+   - platform: watchos
+     scheme: "CareKit"
```
gavirawson-apple, line 9: It looks like SPI automatically infers a configuration based on our repo, can you walk me through why this file is explicitly needed?

gavirawson-apple (commented):
Incredible fixes and improvements here Corey, thanks for all the hard work. And so soon after the release! 

I did a review pass and left some comments inline. I've also created a [small PR](https://github.com/cbaker6/CareKit/pull/7) with some additions to the work you've done here.

.spi.yml
cbaker6: SPI hosts them on their website, though you can keep building the docs on GitHub pages. You can see the ones I made for ParseCareKit (it even tracks betas): https://swiftpackageindex.com/netreconlab/ParseCareKit/0.13.1/documentation/parsecarekit

.spi.yml
cbaker6: It can most of the time, but sometimes it can’t. The latest changes you pushed combined the targets, but it’s possible SPI can still make a mistake because they’re multiple frameworks in one. You can see the builds for CareKit here: https://swiftpackageindex.com/carekit-apple/CareKit/builds

Also, the docs won’t build without being told specifically to do so in the spi file

OCKCatalog/OCKCatalog.xcodeproj/project.xcworkspace/contents.xcworkspacedata
cbaker6: Seems like it should, do you want me to add files like these to the gitignore?

CareKitStore/CareKitStoreTests/OCKHealthKitStore/TestHealthKitPassthroughStore+Events.swift
```diff
@@ -37,7 +37,7 @@ import XCTest
 // Note, we test the event stream and not the outcome stream because the outcome stream
 // calls into the event stream. Testing the outcome stream is unnecessary.
 
-
+@available(iOS 15, watchOS 8, *)
```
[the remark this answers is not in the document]
cbaker6: If I remember correctly, the majority of devices that run on 15 should also run on 13. I need to look up the limitations for watchOS. So you want the lowest phone and watch to be 15 and 8 respectively?

.github/workflows/swift.yml
```diff
@@ -6,21 +6,25 @@ on:
...
     strategy:
       matrix:
-        destination: ['platform=iOS\ Simulator,OS=15.0,name=iPhone\ 11\ Pro\ Max']
-        scheme: ['CareKit\ iOS', 'CareKitStore\ iOS', 'CareKitUI\ iOS', 'CareKitFHIR']
-  
+        destination: ['platform=iOS\ Simulator,OS=16.2,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.1,name=Apple\ Watch\ Series\ 5\ \(40mm\)']
+        scheme: ['CareKit', 'CareKitStore', 'CareKitUI', 'CareKitFHIR']
+        exclude:
+          - destination: 'platform=watchOS\ Simulator,OS=9.1,name=Apple\ Watch\ Series\ 5\ \(40mm\)'
+            scheme: 'CareKitUI'
```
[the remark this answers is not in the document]
cbaker6: This was just all of the tests in the framework are for iOS, there are no current watchOS tests so I skipped it, it builds fine though

CareKit.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
cbaker6: I’m not sure, my Xcode generates these automatically when I build using the workspace file

.spi.yml
cbaker6: You don't have to do anything special to release these other than having the `.spi.yml` file in the repo. I simply create my releases using the standard GitHub release process and SPI adds the documents (usually within ~1 hour of the release)

CareKitStore/CareKitStoreTests/Streaming/TestCoreDataQueryMonitor.swift
```diff
@@ -214,6 +214,7 @@ class TestCoreDataQueryMonitor: XCTestCase {
         XCTAssertEqual(expectedTitles, observedTitles)
     }
 
+    #if compiler(>=5.8.0)
```
cbaker6, line 217: @gavirawson-apple I added this here because GitHub actions currently only supports up to macOS 12 and Xcode 14.2. We can add a note to remove this when actions releases a macOS 13 image. More info here: https://github.com/actions/runner-images/issues/6426#issuecomment-1479974889

My observations over the last couple of years, it takes actions awhile to upgrade to the latest macOS, I suspect this will be June/July

CareKit/CareKit/iOS/Task/View Controllers/OCKSurveyTaskViewController.swift
```diff
@@ -133,65 +133,81 @@ open class OCKSurveyTaskViewController: OCKTaskViewController<OCKSurveyTaskViewS
...
+                warningAlert.addAction(cancelAction)
+                warningAlert.addAction(confirmAction)
+                
+                /*
+                 TODO: Remove in the future. Explicitly setting the tint color here to support
+                 current developers that have a SwiftUI lifecycle app and wrap this view
+                 controller in a `UIViewControllerRepresentable` implementation...Tint color
+                 is not propagated...etc.
+                 */
+                warningAlert.view.tintColor = determineTintColor(from: view)
```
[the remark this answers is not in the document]
cbaker6: > The hope is to move away from wrapping our UIKit views for SwiftUI in part because doing so causes some unintended side effects (card sizing issues, tint color issues, etc.). There are definitely still tools we'll need to add to the framework to make the process even easier, but I'm hoping we can slowly to identify those gaps and build those up together. What do you think of that direction? Should we encourage people down that path and remove patches like this?

I believe this a good idea for the future as I eventually would like to use CareKit and SwiftUI entirely. My concern now is that the SwiftUI cards/views are not close enough in parity with the UIKit views. For example the Contact cards, SurveyTaskView, and WeekHeader. In addition, there's still reliance on asking for permissions, ResearchKit surveys, and opening other OS specific views such as the contact list which all seem to be UIKit based.

When I describe the reasons I use CareKit to build health care apps, I mention that CareKit allows for fast, themeable, attractive, offline-first, distributed based applications. Since most of the UIKit and SwiftUI views do different things, we lose some of the "fast" and "themeable" aspects because developers have to spend time building more custom SwiftUI views to handle what the UIKit cards currently cover and have colors that don't aline with the theme of an app.

In summary, I think the little bit of code to insure the views look the same across apps is a major upside to building current apps with CareKit. As the SwiftUI and UIKit views/cards come closer to parity, I think we can remove

.spi.yml
cbaker6: Let me know if need me to answer more questions from the comment added here: https://github.com/carekit-apple/CareKit/pull/691/files#r1162276983

CareKitStore/CareKitStoreTests/OCKHealthKitStore/TestHealthKitPassthroughStore+Events.swift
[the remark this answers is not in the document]
cbaker6: The other issue that will occur with not having `@available(iOS 15, watchOS 8, *)` is the specification of of this line `Package.swift`: https://github.com/carekit-apple/CareKit/blob/449fd35e97b54974895bb3a7741f8294699d9cd3/Package.swift#L7 

Basically, if a developer opens the project using the Package file, they can't run tests because the SPM file specifies a lower target than the Xcode Project

.spi.yml
cbaker6: Awesome Sven, thanks for looking at this!

CareKit/CareKit.xcodeproj/project.pbxproj
```diff
@@ -883,8 +883,9 @@
 		8605A5B11C4F04EC00DD65FF /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
+				BuildIndependentTargetsInParallel = YES;
```
cbaker6, line 886: This is a new setting Xcode 14.3 automatically adds

CareKit/CareKit.xcodeproj/project.pbxproj
```diff
@@ -1107,6 +1107,7 @@
 		5196C801226F8F8F00F1C2A2 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALLOW_TARGET_PLATFORM_SPECIALIZATION = YES;
```
cbaker6, line 1110: I believe this was added in newer versions of Xcode projects (after the original CareKit project was created) that allows projects to build on arm/intel, etc. Since the original project was built with an old version of Xcode, many of the flags updated when I modified them in Xcode 14.3

CareKit/CareKit.xcodeproj/project.pbxproj
```diff
@@ -1130,16 +1131,19 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = Apple.CareKitTests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SUPPORTED_PLATFORMS = "watchsimulator watchos iphonesimulator iphoneos";
+				SUPPORTS_MACCATALYST = YES;
```
cbaker6, line 1135: When I set the project to build for iOS/watchOS, it automatically added catalyst, I don't see this directly in the Xcode IDE, but I suspect it doesn't hurt much

CareKit/CareKit.xcodeproj/project.pbxproj
```diff
@@ -1130,16 +1131,19 @@
...
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_VERSION = 5.0;
-				TARGETED_DEVICE_FAMILY = "1,2";
+				TARGETED_DEVICE_FAMILY = "1,2,4";
```
cbaker6, line 1139: The `4` is added to support watchOS

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
cbaker6: I'm looking further into this as I know how to replicate it, but if I set breakpoints and step through them. This leads to believe that it could be related to how I'm setting up the Stores or there could be a threading issue with the streaming store. I only get the crash on Xcode 14.3, it doesn't happen on Xcode 14.2

cbaker6:
One additional observation I've noticed is that if you add two `@CareStoreFetchRequest` to a view, the view will crash. For example if I want to tie a custom card to `OCKPatient` and another to `OCKContact`, but show them on same screen. This code inside the same view will crash:

```swift
@CareStoreFetchRequest(query: OCKPatientQuery(for: Date())) private var patients
@CareStoreFetchRequest(query: OCKContactQuery(for: Date())) private var contacts
```

The workaround isn't hard, just only use 1 `@CareStoreFetchRequest` per view and place the views in a `VStack` or other container. I just didn't expect it to crash the app when originally tested the code

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
cbaker6: This is where I get the crash the most in Xcode 14.3 which is why I originally thought it had something to do with `Calendar.current`: 
image

The call stack shows some of the streaming going on. In addition, this is in the console window:

```bash
2023-04-11 17:53:16.229604-0400 OCKSample[50867:4961092] [CareKit] Beginning to stream data from the store
2023-04-11 17:53:16.259117-0400 OCKSample[50867:4961092] [CareKit] Beginning to stream data from the store
2023-04-11 17:53:16.592551-0400 OCKSample[50867:4961396] [error] error: NULL _cd_rawData but the object is not being turned into a fault
CoreData: error: NULL _cd_rawData but the object is not being turned into a fault
(lldb) 
```

If I set a breakpoint here, `po` a couple properties, then run, it won't crash

cbaker6: > Nice catch. I wonder if a good solution here is to pass in the calendar when it's needed and store it elsewhere (maybe the OCKStore?).

This sounds like the right solution to me as I don’t see a reason to send it to the server

CareKit/CareKit/Shared/Synchronization/CareStoreFetchRequestController.swift
```diff
@@ -92,6 +92,10 @@ final class CareStoreFetchRequestController<Result, Query>: ObservableObject {
         )
     }
 
+    deinit {
+        cancelStreamingTask()
```
cbaker6, line 96: Calling the already created method to ensure streams shut down gracefully

CareKitStore/CareKitStore/StoreCoordinator/OCKStoreCoordinator.swift
```diff
@@ -208,7 +208,7 @@ open class OCKStoreCoordinator: OCKAnyStoreProtocol {
         // a single combined result is sent through the stream.
 
         // Create a stream that outputs a single empty result
-        let initialResults: AsyncLazySequence<[[T]]> = [[]].async
+        let initialResults: AsyncSyncSequence<[[T]]> = [[]].async
```
cbaker6, line 211: All of these are related to 0.1.0 changes in https://github.com/apple/swift-async-algorithms/pull/246

CareKitUI/CareKitUITests/TestColorExtension.swift
```diff
@@ -57,8 +56,7 @@ class TestColorExtension: XCTestCase {
 
     func testClamping() {
         let start = UIColor(red: 0, green: 0, blue: 0, alpha: 1)
-        let lightened = start.lightened(10)
+        let lightened = start.lightened(1)
```
cbaker6, line 59: This has a pre-condition where the argument passed to `lightened` has to be `<=1`. For some reason it hasn't been failing on the iOS tests, but the watchOS tests caught the error.

cbaker6:
@gavirawson-apple this is ready for review. You can see the passing tests in the CI here: https://github.com/cbaker6/CareKit/pull/8

The CI should begin running against this repo once it's approved

CareKitUI/CareKitUITests/TestColorExtension.swift
cbaker6: I found the issue. The reason was `TestColorExtension` wasn't included in the XCTest scheme so it never ran when the CI tests the Xcode project. I noticed the issue in SPM after https://github.com/carekit-apple/CareKit/pull/691#discussion_r1172662798

The update is here: https://github.com/carekit-apple/CareKit/pull/691/commits/225341b1785513785b7d4a9d4538c2dd48bc0554

CareKitStore/CareKitStore/CoreData/Synchronization/OCKStore+Synchronization.swift
```diff
@@ -215,8 +215,13 @@ extension OCKStore: OCKRemoteSynchronizationDelegate {
                                 self.context.knowledgeVector.increment(
                                     clockFor: self.context.clockID
                                 )
+                                
+                                // 8. Lock in the changes. If this fails, all
+                                //    merged changes will be rolled back and
+                                //    we'll need to try again later.
+                                try self.context.save()
```
cbaker6, line 222: This bug wasn't easy to find, but was introduced with the changes in #553. The is is the last increment to the knowledge vector is never saved in context, causing data to be lost when saving OCK objects using a CareKit card or other view because it doesn't have the most up-to-date clock value.

The current test cases don't catch this issue because they are testing the stores directly as oppose to using the CareKit cards to save the outcomes.

The example case to replicate the issue is below:

1. Setup `Device1` to connect to a `OCKRemoteServer`
2. Add tasks
3. Save an outcome using a CareKit card and sync `Device1` to the `OCKRemoteServer`
4. Bring a new `Device2` online and sync it to the `OCKRemoteServer`. `Device2` will now reflect everything on `Device1`
5. Delete the saved outcome using a CareKit card on `Device2`. This will trigger a pullRevision, but notice during pushRevisions no new revision records will be found because the CareKit didn't have the correct Knowledge Vector clock value because it wasn't saved in context. Causing `Device1` to never receive the deleted outcome, forcing the devices out-of-sync

cbaker6: This can also be done after push revisions is complete, but I'm not sure if it's actually better and requires extra code:

```swift
// 4. Detect and resolve any conflicts that exist. The store
                    //    is a CRDT, so there aren't any conflicts at the data
                    //    layer, but we want to resolve concurrent changes.
                    self.resolveConflicts { error in

                        self.context.perform {
                            do {
                                // 5. Lock in the changes. If this fails, all
                                //    merged changes will be rolled back and
                                //    we'll need to try again later.
                                try self.context.save()

                                // 6. Local revisions will now include any patches
                                //    that were applied during conflict resolution.
                                let localKnowledge = self.context.knowledgeVector

                                let localRevisions = try self.computeRevisions(
                                    since: remoteKnowledge
                                )

                                // 7. Bump knowledge vector to indicate that any
                                //    objects created beyond this point in time
                                //    are considered unknown to the peer.
                                self.context.knowledgeVector.increment(
                                    clockFor: self.context.clockID
                                )

                                // 8. Push conflict resolutions + local changes to remote
                                remote.pushRevisions(
                                    deviceRevisions: localRevisions,
                                    deviceKnowledge: localKnowledge) { error in

                                    if let error = error {
                                        os_log("Failed to push revision. %{private}@",
                                               log: .store, type: .error, error as NSError)
                                    }
                                    
                                    self.context.perform {
                                        do {
                                            // 9. Lock in the changes. If this fails, all
                                            //    merged changes will be rolled back and
                                            //    we'll need to try again later.
                                            try self.context.save()
                                        } catch {
                                            self.context.rollback()
                                            completion(error)
                                            return
                                        }
                                        // 10. The sync is still considered successful
                                        //    even if the remote doesn't accept the
                                        //    push. The next time we sync with it, it
                                        //    will still have the same knowledge
                                        //    vector, which means the proper diff will
                                        //    be generated.
                                        completion(nil)
                                    }
                                }
                            } catch {
                                self.context.rollback()
                                completion(error)
                            }
                        }
                    }
```

gavirawson-apple: Nice catch!

CareKitStore/CareKitStore/Structs/OCKScheduleElement.swift
```diff
@@ -88,7 +88,9 @@ public struct OCKScheduleElement: Codable, Equatable {
...
+    private static var calendar: Calendar = {
+        Calendar.current
+    }()
```
gavirawson-apple, line 93: Let's avoid storing the default calendar here in case it changes after this schedule element is created. (Though we likely need some other improvements here to handle time zone changes)

```suggestion
    private var calendar: Calendar { .current }
```

gavirawson-apple (approved):
Looks amazing Corey, I just have one minor suggestion to consider before we merge away.

cbaker6:
@gavirawson-apple let me know if you need anything else from me on this PR

gavirawson-apple:
Thanks Corey, everything is set here. We're just going through some process on our end.

PR #696, opened 2023-05-04
Build CareKitStore and CareKitFHIR for macOS

cbaker6:
- [x] Build CareKitStore and CareKitFHIR for macOS 13+
- [x] Fix documentation build on SPI
- [x] Bump CI to macOS 13 and Xcode 14.3

## Notes
- Cannot build CareKit due to CareKitUI using UIKit, UIColor, etc. This seems like it would need a lot more changes. I speculate with future changes, the SwiftUI based CareKit views build for macOS
- HealthKit is only available on macOS 13+
- ~~Can't build/test in GitHub actions due to macOS 12 only being available, so I removed the changes in the CI for now. I have tested locally, and all tests pass for CareKitStore and CareKitFHIR on macOS 13~~

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -15,13 +15,8 @@ jobs:
     runs-on: macos-12
     strategy:
       matrix:
-        destination: ['platform=iOS\ Simulator,OS=16.2,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.1,name=Apple\ Watch\ Series\ 5\ \(40mm\)', platform=macOS]
+        destination: ['platform=iOS\ Simulator,OS=16.2,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.1,name=Apple\ Watch\ Series\ 5\ \(40mm\)']
         scheme: ['CareKit', 'CareKitStore', 'CareKitUI', 'CareKitFHIR']
-        exclude:
-          - destination: 'platform=macOS'
-            scheme: 'CareKit' 
-          - destination: 'platform=macOS'
-            scheme: 'CareKitUI' 
     steps:
       - uses: actions/checkout@v2
       - name: Set Xcode Version
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: macos-13
     strategy:
       matrix:
-        destination: ['platform=iOS\ Simulator,OS=16.3,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.2,name=Apple\ Watch\ Series\ 5\ \(40mm\)', 'platform=macOS']
+        destination: ['platform=iOS\ Simulator,OS=16.4,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.4,name=Apple\ Watch\ Series\ 5\ \(40mm\)', 'platform=macOS']
         scheme: ['CareKit', 'CareKitStore', 'CareKitUI', 'CareKitFHIR']
         exclude:
           - destination: 'platform=macOS'
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -12,15 +12,20 @@ concurrency:
 
 jobs:        
   test:
-    runs-on: macos-12
+    runs-on: macos-13
     strategy:
       matrix:
-        destination: ['platform=iOS\ Simulator,OS=16.2,name=iPhone\ 14\ Pro\ Max', 'platform=watchOS\ Simulator,OS=9.1,name=Apple\ Watch\ Series\ 5\ \(40mm\)']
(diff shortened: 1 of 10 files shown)


cbaker6:
Hey @finestructure, for some reason the docs didn't build with the current config: https://github.com/carekit-apple/CareKit/blob/b07289557fa0850a351aaebc0bbc4d7209abe580/.spi.yml#L1-L11

If you get a chance, can you let me know if there's a build log for the docs and let me know if you see a quick fix?

finestructure:
Hi @cbaker6 , I'm not at a computer right now but looking at your manifest file I'm pretty sure I know what the problem is. The second ios config overrides the first. You'll want to combine them:

```
version: 1
builder:
   configs:
   - documentation_targets: ["CareKit", "CareKitUI", "CareKitStore", "CareKitFHIR"]
     platform: ios
     scheme: "CareKit"
   - platform: watchos
     scheme: "CareKit"
   - platform: macos
     scheme: "CareKit" 
```

cbaker6:
> The second ios config overrides the first. You'll want to combine them

this makes sense, I’ve updated the spi file with your suggestions.

cbaker6:
@gavirawson-apple I believe this is ready for review

finestructure:
@cbaker6 I've looked into the logs and it actually tries to generate the docs as expected, the config isn't the issue it seems.

There are errors like

```
Details:  Error deleting scheme: Cannot modify data because the process disallows saving.
Object:   <IDERunContextManager: 0x60000038c240>
Method:   -deleteRunContexts:completionQueue:completionBlock:
Thread:   <_NSMainThread: 0x6000024a8200>{number = 1, name = main}
Please file a bug at https://feedbackassistant.apple.com with this warning message and any useful information you can provide.
2023-05-03 12:06:52.179 xcodebuild[31287:537047] could not delete old scheme: Error Domain=DVTCustomDataStoringErrors Code=0 "Cannot modify data because the process disallows saving." UserInfo={NSLocalizedDescription=Cannot modify data because the process disallows saving.}
2023-05-03 12:06:52.179 xcodebuild[31287:537047] [MT] DVTAssertions: Warning in /System/Volumes/Data/SWE/Apps/DT/BuildRoots/BuildRoot7/ActiveBuildRoot/Library/Caches/com.apple.xbs/Sources/IDEFrameworks/IDEFrameworks-21812/IDEFoundation/Execution/RunContextManager/IDERunContextManager.m:842
Details:  Error deleting scheme: Cannot modify data because the process disallows saving.
```

You can find the full logs here: https://spi-build-logs.s3.us-east-2.amazonaws.com/66595795-8511-4968-8924-1C743506E315.log

Search for `Generating docs at path` to find the start of the doc generation.

When building docs on ios, we run

```
xcodebuild docbuild -scheme $DOC_TARGET -destination 'generic/platform=ios'
```

followed by

```
xcrun docc process-archive $ARCHIVE_PATH transform-for-static-hosting --output-path $OUTPUT_PATH --hosting-base-path $HOSTING_BASE_PATH
```

where `ARCHIVE_PATH` is the path to the DocC archive generated in the first step. (I don't think it gets to this step, I believe the error is happening during `docbuild`.)

finestructure:
Ok, so the config _is_ in fact the issue but not quite in the way I thought.

I've reproduced the exact issue locally with the original `.spi.yml` and the docs are successfully generated with the proposed changes to the manifest file.

What seems to be happening is that due to the config issue, the _build_ of the package on iOS fails (it tries to build `CareKit-Package`, because no custom `scheme` is set) and although the doc gen step also runs a build, it apparently falls over in a weird way due to the previous build error.

cbaker6:
@finestructure When I execute:

```bash
xcodebuild docbuild -scheme CareKit -destination generic/platform=ios
```

It's successful: 

```bash
** BUILD DOCUMENTATION SUCCEEDED **
```

I then execute:

```bash
xcrun docc process-archive transform-for-static-hosting /Users/x/Documents/Xcode_Projects/CareKit/DerivedData/CKWorkspace/Build/Products/Debug-iphoneos/CareKit.doccarchive --output-path ./test/ --hosting-base-path ""
```

And it produces the correct file structure to host the docs in `/test`

cbaker6:
Looking through https://github.com/carekit-apple/CareKit/pull/696#issuecomment-1536279431 now, it looks like we posted at the same time.

finestructure:
Hah, posted at the same time :) That test is actually incomplete. It does work on its own but falls over if it's running on the back of a failed build. I haven't dug deeper into why that is - it should just all shake itself out with the updated manifest in place!

cbaker6:
> What seems to be happening is that due to the config issue, the build of the package on iOS fails (it tries to build CareKit-Package, because no custom scheme is set) and although the doc gen step also runs a build, it apparently falls over in a weird way due to the previous build error.

I see, do you think the updated config will fix the issue?

```bash
version: 1
builder:
   configs:
   - documentation_targets: ["CareKit", "CareKitUI", "CareKitStore", "CareKitFHIR"]
     platform: ios
     scheme: "CareKit"
   - platform: watchos
     scheme: "CareKit"
   - platform: macos
     scheme: "CareKit"
```

cbaker6:
@finestructure I really appreciate you looking into this!

PR #711, opened 2024-02-29
feat: Update OCKSurveyTaskViewController to support ResearchKit 3.0

cbaker6:
Some minor changes are required to support RK 3.0. 
- https://github.com/carekit-apple/CareKit/pull/711/commits/c9ab6388ce592896c74275cae830654b6035804e Supports both old and new RK, but has redundant code
- https://github.com/carekit-apple/CareKit/pull/711/commits/d1a87996001487789f13cdc3d9bd68101f5a8b28 These changes will make CK not compatible with older versions of RK and doesn't have redundant code

I've left the current commit with the redundant code, but have tested both. The former works with old/new versions of RK. The second only works with new versions of RK

### DIFF
--- a/CareKit/CareKit/iOS/Task/View Controllers/OCKSurveyTaskViewController.swift
+++ b/CareKit/CareKit/iOS/Task/View Controllers/OCKSurveyTaskViewController.swift
@@ -34,6 +34,7 @@ import CareKitUI
 import ResearchKit
 #if canImport(ResearchKitUI)
 import ResearchKitUI
+public typealias ORKTaskViewControllerFinishReason = ORKTaskFinishReason
 #endif
 import UIKit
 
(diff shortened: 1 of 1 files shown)


cbaker6:
If https://github.com/ResearchKit/ResearchKit/pull/1566 is accepted, we can revert this back to https://github.com/carekit-apple/CareKit/commit/d1a87996001487789f13cdc3d9bd68101f5a8b28

CareKit/CareKit/iOS/Task/View Controllers/OCKSurveyTaskViewController.swift
```diff
@@ -32,31 +32,58 @@
 import CareKitStore
 import CareKitUI
 import ResearchKit
+#if canImport(ResearchKitUI)
+import ResearchKitUI
+#endif
```
ronzilla-apple, line 37: Would it be possible to change the entire diff to something like

```suggestion
#if canImport(ResearchKitUI)
import ResearchKitUI
public typealias ORKTaskViewControllerFinishReason = ORKTaskFinishReason
#endif
```

ronzilla-apple (changes requested):
I was able to get a sample project building with only the conditional import and the typealias.
That change worked with both RK3.0 and RK2.x

gavirawson-apple (approved):

gavirawson-apple:
Thanks for the fix Corey!

PR #715, opened 2024-06-23
fix: Update swift-async-algorithms to 1.0.0

cbaker6:
swift-async-algorithms is currently not on the latest version


gavirawson-apple:
Thanks Corey, this is a great update. Because AsyncAlgorithms is heavily used in the framework to stream care store data, give us some time to test the package update and make sure it doesn't break anything. We'll get this merged soon.

gavirawson-apple (approved):

Issue #357, opened 2020-02-10
Updating OCKContact, OCKPatient, etc. creates duplicates in CareStore


cbaker6:
After adding a new contact or patient to the CareStore, calling an update method such as "store.updateContact" creates an additional contact in the CareStore. If I'm understanding the documentation correctly, this should simply update the current contact and create a new version (allowing you go back to the older version if needed).

The problem is that querying contacts then returns each contact as a new one with the same id.

Is if you make 10 updates to contact, such as updates to address, etc. you will end up with 11 different contacts.

If tested this on OCKPatient and the same thing is happening.

The version of CareKit is the latest commit: 78a6ce5e34a58830dd71034cdd87ff1ce7e5c561

cbaker6:
My assumption for a workaround is to use the contact with the latest version or probably the one with the newest "effectiveDate"

erik-apple:
Each time you update one of the versioned entities (patent, care plan, contact, and task), a new entry is created in the table, and the old and new versions are linked together. The intended behavior is that when you query contacts in an interval, the store should return only the *newest* version existing in the given interval. 

You've discovered a bug in which not passing any date results in all versions being returned.

We'll prioritize fixing this bug first as it will affect more users and doesn't have as simple of a work around as the other bug you found for us. You're assumed workaround is probably the best option available until we can issue a patch.

We appreciate your help ironing out these bugs!

erik-apple:
We did some work to fix this, but upon further reflection, are now leaning toward saying that the present behavior is correct, and that if you only want the latest version, you need to specify a date on the query.

Consider a case in which a patient is taking a medication, and the dose for that medication is periodically updated. Now let's imagine that we'd like to create a view controller which shows the prescription history for that medication. In order to do that, we need to be able to query *all versions* of the associated `OCKTask`. If queries only return the latest version of the task, then there is no way to gather all past versions without knowing a-priori when updates were made.

The present behavior allows you to ask for all versions by not passing a date, or to get the valid version in a date window.

```swift
// This query will return all versions, past and present, of the task "doxylamine"
var query = OCKTaskQuery()
query.ids = ["doxylamine"]

// This query will return the newest version of the task as of today
var query = OCKTaskQuery(for: Date())
query.ids = ["doxylamine"]

// This query will return the current version of the task as it would have appeared a week ago
var query = OCKTaskQuery(for: aWeekAgo)
query.ids = ["doxylamine"]
```

[...]

PR #718, opened 2024-07-07
fix: KeyPath crash when querying CareStore

cbaker6:
The framework crashes when you stream using: `OCKOutcomeQuery`, `OCKCarePlanQuery`, `OCKContactQuery`, `OCKPatientQuery`, `OCKTaskQuery`. Most likely happens with a standard fetch as well, but I didn't test those...

Resulting in:

image

image

This can be replicated by using a simple streaming query within a view:

```swift
// A view that has the CareStore in it's environment 
struct SimpleCareView: View {

    @CareStoreFetchRequest(query: query()) private var outcomes

    var body: some View {
         // I want to display outcomes...
    }

    // Simple query
    static func query() -> OCKOutcomeQuery {
        OCKOutcomeQuery(for: Date())
    }
}
```

This most likely results from `NSSortDescriptor` not liking keyPaths of value types and prefers reference types (I'm guessing as I'm seen some other issues in the past with value type keyPaths and using NS...). In addition, since we are querying from CoreData, we should probably use the CD representations. Lastly, all other uses of `NSSortDescriptor` in `CareKitStore` use their `OCKCD...` counterparts instead of the value type counterparts. 

## Note

When I ran the test suite, none of the current tests for the framework hit the lines I changed which is probably why this wasn't caught 🙃. This is because the tests are conducted at a lower level than the value type representations, but solidifies that we should use the CoreData Entity representations for querying keyPaths: https://github.com/carekit-apple/CareKit/blob/7416cce107875e58d0dbfd54afb16e007ef79033/CareKitStore/CareKitStoreTests/Streaming/TestCoreDataQueryMonitor.swift#L65-L70

https://github.com/carekit-apple/CareKit/blob/7416cce107875e58d0dbfd54afb16e007ef79033/CareKitStore/CareKitStore/Streaming/CoreDataQueryMonitor.swift#L47-L63

### DIFF
--- a/CareKitStore/CareKitStore/StoreCoordinator/OCKPersistentStoreCoordinator+Outcomes.swift
+++ b/CareKitStore/CareKitStore/StoreCoordinator/OCKPersistentStoreCoordinator+Outcomes.swift
@@ -45,7 +45,7 @@ extension OCKStoreCoordinator {
         }
 
         let sortDescriptor = NSSortDescriptor(
-            keyPath: \OCKAnyOutcome.id,
+            keyPath: \OCKCDOutcome.id,
             ascending: true
         )
 
--- a/CareKitStore/CareKitStore/Structs/Queries/OCKCarePlanQuery.swift
+++ b/CareKitStore/CareKitStore/Structs/Queries/OCKCarePlanQuery.swift
@@ -42,9 +42,9 @@ public struct OCKCarePlanQuery: Equatable, OCKQueryProtocol {
         var nsSortDescriptor: NSSortDescriptor {
             switch self {
             case let .effectiveDate(ascending):
-                return NSSortDescriptor(keyPath: \OCKAnyCarePlan.effectiveDate, ascending: ascending)
+                return NSSortDescriptor(keyPath: \OCKCDCarePlan.effectiveDate, ascending: ascending)
             case let .title(ascending):
-                return NSSortDescriptor(keyPath: \OCKAnyCarePlan.title, ascending: ascending)
+                return NSSortDescriptor(keyPath: \OCKCDCarePlan.title, ascending: ascending)
             }
         }
     }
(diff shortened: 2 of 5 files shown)


CareKitStore/CareKitStore/StoreCoordinator/OCKPersistentStoreCoordinator+Outcomes.swift
```diff
@@ -45,7 +45,7 @@ extension OCKStoreCoordinator {
         }
 
         let sortDescriptor = NSSortDescriptor(
-            keyPath: \OCKAnyOutcome.id,
+            keyPath: \OCKCDOutcome.id,
```
cbaker6, line 48: I have some thoughts here...

It looks like the public sort descriptor for `OCKOutcomeQuery` was removed in 3.0. My guess is the code above is here to maintain some consistency when the same query is made when the exact same data is in the database?

I think there is at least one case (maybe two) to add the public SortDescriptor back. In #553 `OCKOutcome` was turned into a versioned entity, but querying for `OCKOutcome` doesn't necessarily follow the [comments](https://github.com/carekit-apple/CareKit/issues/357#issuecomment-584334257) and [documentation](https://github.com/carekit-apple/CareKit?tab=readme-ov-file#carekitstore-) which can make since due to the nature of an outcome (may want to know the result for a particular day as oppose to the latest version). Still, there are times where a developer will want results by `effectiveDate`, which all current queries to versioned objects currently allow, except for `OCKOutcome`. Because of this, I've added extensions in my helper framework to help handle `OCKOutcome` results https://github.com/netreconlab/CareKitEssentials/pull/12

If you agree, we can add the public SortDescriptor back, but include `effectiveDate` like: https://github.com/carekit-apple/CareKit/blob/7416cce107875e58d0dbfd54afb16e007ef79033/CareKitStore/CareKitStore/Structs/Queries/OCKTaskQuery.swift#L36-L64

Another case can be added to the SortDescriptor for `taskOccurrenceIndex`, which could be useful. We can then add the case from sorting by `id` (the above code), but mark that one as internal as I can imagine this could over time. We can prepend the sort by `id` to all queries to replicate the current behavior of the above block of code.

Let me know what you think and I can create a follow-up PR after a decision is made on this one.

PR #721, opened 2024-11-28
fix: App crash when faults are not realized on CoreData objects

cbaker6:
- [x] Sometimes the CoreData objects cross thread boundaries, so this ensures faults are realized for the rest of the objects that may be used in other places


gavirawson-apple (approved):

gavirawson-apple:
Thanks for the hard work Corey! We'll be merging a few of your fixes now.

PR #724, opened 2025-03-02
fix: HealthStore deleteOutcomes should return on success

cbaker6:
- [x] Ensure HealthStore returns on correct queue after deletion


gavirawson-apple (approved):
Great fix Corey

PR #728, opened 2025-09-01
Fix duplicate outcome validation during sync

beltradini:
This pull request adds a regression test and a code fix to ensure that duplicate outcome validation in the CareKit store works correctly when outcomes have newer versions. The test verifies that outcomes with a next version skip duplicate validation, and the code fix implements this logic in the validation method.

**Testing improvements:**

* Added a new test case in `TestOutcomeDuplicate.swift` to verify that outcomes with a next version do not trigger duplicate validation, and that outcomes without a next version correctly throw an error when duplicates exist.

**Bug fix:**

* Updated the `validateForInsert()` method in `OCKCDOutcome.swift` to skip duplicate validation if the outcome has a next version, preventing false positives when validating outcomes that are part of a version chain.

### DIFF
--- a/CareKit/CareKitTests/Synchronization/TestOutcomeDuplicate.swift
+++ b/CareKit/CareKitTests/Synchronization/TestOutcomeDuplicate.swift
@@ -1,0 +1,74 @@
+/*
+ Copyright (c) 2016-2025, Apple Inc. All rights reserved.
+ 
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ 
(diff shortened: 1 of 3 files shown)


gavirawson-apple:
Thanks @beltradini! I may be thinking about this wrong, but isn't it important we prevent a client from updating an old version of an outcome? It's been some time since I've toyed with the versioning system here, but I believe a client should only be able to update the latest version of an object to ensure the version chain remains accurate. If you have a specific situation in mind where a client needs to update an old version of an outcome that might clear this up for me.

PR #737, opened 2025-12-14
fix: Add more Hashable conformance

### LINKED ISSUE #726: Update many of the concrete CareKit types to Hashable
Many of the types in CareKit are structs that conform to `Equatable`, but not `Hashable`. This causes issues when using the respective types to update a SwiftUI view. See a bug that is mentioned here for reference https://github.com/netreconlab/CareKitSample-ParseCareKit/pull/145#issuecomment-2863932587

Can types such as OCKAnyEvent, OCKPatient, OCKOutcome, etc. be made `Hashable`. This is harder than it seems as it will initially break a lot of the CareKit code.

Also, a nice to have is to have `@CareStoreFetchRequest` made `Hashable` and to be able to get return types of [OCKPatient], [OCKCarePlan], etc. Note, I'm not suggestion that the protocols `OCKAnyPatient` conform to `Hashable` as that will cause additional breaks in code.

A workaround along with conformance for many of the types are in our Essentials framework (the files typically end in `...+Hashable.swift`): https://github.com/netreconlab/CareKitEssentials/tree/main/Sources/CareKitEssentials/Extensions

cbaker6:
Close #726

Continuation on #736

### DIFF
--- a/CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
+++ b/CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
@@ -31,7 +31,7 @@
 import Foundation
 
 /// Conforming a type to `OCKAnyTask` allows it to be queried and displayed by CareKit.
-public protocol OCKAnyTask: Identifiable, Sendable {
+public protocol OCKAnyTask: Sendable {
 
     /// A user-defined unique identifier, typically human readable.
     var id: String { get }
--- a/CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
+++ b/CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
@@ -31,7 +31,7 @@
 import Foundation
 
 /// Conforming a type to `OCKAnyTask` allows it to be queried and displayed by CareKit.
-public protocol OCKAnyTask: Sendable {
+public protocol OCKAnyTask: Identifiable, Sendable {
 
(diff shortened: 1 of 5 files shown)


CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
```diff
@@ -31,7 +31,7 @@
 import Foundation
 
 /// Conforming a type to `OCKAnyTask` allows it to be queried and displayed by CareKit.
-public protocol OCKAnyTask: Sendable {
+public protocol OCKAnyTask: Identifiable, Sendable {
```
gavirawson-apple: We have long-standing problem ever since the `Identifiable` was introduced - our `id` property isn't a stable ID for different versions of the same task, but our `uuid` is. I think that means we can't conform `OCKAnyTask` to `Identifiable` directly, the best you can do right now is wrap `OCKAnyTask` in an `Identifiable` type.

gavirawson-apple (changes requested):
Looking great, `Hashable `conformance will make handling these data structures much easier. I think we just need to remove the `Identifiable` conformance before merging. Let's brainstorm the best way to move towards `Identifiable` conformance without negatively impacting developers.

CareKitStore/CareKitStore/Protocols/Tasks/OCKAnyTask.swift
cbaker6: I see, I’ll make an update soon.

PR #739, opened 2026-01-06
feat: Add more Sendable arguments to remote synchronizer

cbaker6:


gavirawson-apple (approved):
Another nice catch, thanks Corey.

PR #741, opened 2026-01-11
fix: Change README badge to Xcode 16.3

cbaker6:
Change badge to lowest Xcode that supports Swift 6.1.


gavirawson-apple (approved):
Thanks for catching this Corey!

